PR #593, opened 2020-07-16
Import connectivity to new region sidebar

fsdavid:


src/services/state/viewerState.store.ts
```diff
@@ -65,7 +66,7 @@ export const defaultState: StateInterface = {
   parcellationSelected: null,
   templateSelected: null,
   connectivityRegion: '',
-
+  connectivityVisible: false,
```
xgui3783, line 69: I think instead of having `connectivityVisible` state, we should have a more generic `overwrittenColorMap` property in `ngViewerStateStore`

src/ui/connectivityBrowser/connectivityBrowser.template.html
```diff
@@ -10,11 +10,12 @@
             show-toolbar="false"
             show-description="false"
             show-dataset-name="false"
-            custom-dataset-selector="true"
-            [customHeight]="componentHeight + 'px'"
-            [ngStyle]="connectivityComponentStyle">
+            custom-dataset-selector="true">
+<!--        [customHeight]="componentHeight + 'px'"-->
+<!--        [ngStyle]="connectivityComponentStyle">-->
```
xgui3783, line 15: can you remove these if not using?

src/ui/connectivityBrowser/connectivityBrowser.template.html
```diff
@@ -10,11 +10,12 @@
...
         <div slot="header" class="w-100 d-flex justify-content-between mt-3">
-            <button (click)="closeConnectivityView(); setDefaultMap()" class="w-100" mat-button><i class="far fa-times-circle cursor-pointer"></i> Close connectivity</button>
+<!--            <button (click)="closeConnectivityView(); setDefaultMap()" class="w-100" mat-button><i class="far fa-times-circle cursor-pointer"></i> Close connectivity</button>-->
```
xgui3783, line 17: ditto, remove if not using?

src/ui/nehubaContainer/nehubaContainer.component.ts
```diff
@@ -163,6 +164,8 @@ export class NehubaContainer implements OnInit, OnChanges, OnDestroy {
 
   public selectedParcellation: any | null
 
+  public selectedParcellationHasConnectivity: boolean = false
```
xgui3783, line 167: I would rather not introduce a property, and instead parse it in template instead

xgui3783:
also, I thought you will add some contents here instead:

`https://github.com/HumanBrainProject/interactive-viewer/pull/593/files#diff-bcd50078d4f1166cf85fcd9aa24b1401L280`

fsdavid:
> also, I thought you will add some contents here instead:
> 
> https://github.com/HumanBrainProject/interactive-viewer/pull/593/files#diff-bcd50078d4f1166cf85fcd9aa24b1401L280

Sorry, I could not catch what do you mean

xgui3783:
right now, if no region is selected, this is what it looks like:

Screenshot from 2020-07-16 17-06-57

xgui3783:
Actually I just had another thought about it, don't worry about either comments for now.

We will refactor later.

I will merge this PR, so we can work on other things

PR #621, opened 2020-08-02
add connectivity to region chip

fsdavid:


xgui3783:
can you fix the PR? I think it has something to do with private/public properties

fsdavid:
> can you fix the PR? I think it has something to do with private/public properties

Sure, Fixed.

PR #635, opened 2020-08-24
Automatically apply connectivity colors

fsdavid:


src/ui/databrowserModule/preview/previewDatasetFile.directive.ts
```diff
@@ -103,7 +109,11 @@ export class PreviewDatasetFile extends PreviewBase implements OnChanges{
       this.snackbar.open(`Cannot preview dataset file. Neither file nor filename are defined.`)
       return
     }
+    this.store$.dispatch(ngViewerActionClearView({
+      payload: {connectivity: true}
+    }))
```
xgui3783, line 114: There should be no reason you need to modify previewdatasetfile.directive.ts

src/ui/connectivityBrowser/connectivityBrowser.component.ts
```diff
@@ -200,19 +199,19 @@ export class ConnectivityBrowserComponent implements AfterViewInit, OnDestroy, A
...
-      if (this.defaultColorMap) this.setDefaultMap()
+    public closeConnectivityView(setDefault = true) {
+      if (this.defaultColorMap && setDefault) this.setDefaultMap()
       this.closeConnectivity.emit()
-      this.store$.dispatch({
-        type: CLEAR_CONNECTIVITY_REGION,
-      })
+      // this.store$.dispatch({
+      //   type: CLEAR_CONNECTIVITY_REGION,
+      // })
```
xgui3783, line 207: can you remove instead of comment?

src/ui/connectivityBrowser/connectivityBrowser.component.ts
```diff
@@ -146,11 +142,14 @@ export class ConnectivityBrowserComponent implements AfterViewInit, OnDestroy, A
               // ToDo Fix in future to use component
               const a = document.querySelector('hbp-connectivity-matrix-row')
               a.downloadCSV()
-            } else if (e.detail.name === 'Apply colors to viewer') {
-              this.defaultColorMap && this.toggleConnectivityOnViewer( {checked: this.showConnectivityToggle? false : true})
             }
+            // else if (e.detail.name === 'Apply colors to viewer') {
+            //   this.defaultColorMap && this.toggleConnectivityOnViewer( {checked: !this.showConnectivityToggle})
+            // }
```
xgui3783, line 148: can you remove instead of comment?

src/ui/connectivityBrowser/connectivityBrowser.component.ts
```diff
@@ -89,35 +87,33 @@ export class ConnectivityBrowserComponent implements AfterViewInit, OnDestroy, A
...
+            })
           }
         }),
       )
-      this.subscriptions.push(this.templateSelected$.subscribe(t => {
-        this.closeConnectivityView()
-      }))
+      // this.subscriptions.push(this.templateSelected$.subscribe(t => {
+      //   this.closeConnectivityView(false)
+      // }))
```
xgui3783, line 116: can you remove instead of coment?

src/ui/connectivityBrowser/connectivityBrowser.component.ts
```diff
@@ -38,10 +39,7 @@ export class ConnectivityBrowserComponent implements AfterViewInit, OnDestroy, A
     public math = Math
 
     public connectivityComponentStyle: any = {
-      // ToDo - new popup panels from layer container hardcoded height and when I'm trying to grow connectivity popup, it overflows below.
-      //  For that I will just set hardcoded height, If we will able to grow new popups dynamically, below commented maxHeight should be apply.
-      // maxHeight: +(+getWindow().innerHeight - 200) + 'px',
-      maxHeight: '500px'
+      maxHeight: +(+getWindow().innerHeight - 200) + 'px',
```
xgui3783, line 42: I will have to double check but using innerHeight could be a performance killer

src/services/state/ngViewerState.store.ts
```diff
@@ -425,6 +448,17 @@ export class NgViewerUseEffect implements OnDestroy {
...
+    this.cleanView$ = this.actions.pipe(
+      ofType(ngViewerActionClearView.type),
+      mergeMap((a) => {
+        if (a['payload']['connectivity']) {
+          return [{type: CLEAR_CONNECTIVITY_REGION},
+            {type: SET_CONNECTIVITY_VISIBLE, payload: null}]
+        }
+        return EMPTY
+      })
+    )
```
xgui3783, line 461: I think you misunderstood me. 

when you want to clear pmap, you call the method `ngViewerActionClearView` with payload. I will implement the logic, because there are a lot of plumbing required.

So in the case of connectivity browser, when user clicks `show connectivity heat map`, you would dispatch the action `ngViewerActionClearView`

src/services/state/ngViewerState.store.ts
```diff
@@ -425,6 +448,17 @@ export class NgViewerUseEffect implements OnDestroy {
         }
       }),
     )
+
+    this.cleanView$ = this.actions.pipe(
+      ofType(ngViewerActionClearView.type),
+      mergeMap((a) => {
```
xgui3783, line 454: separately, I think this should be just map operator (or filter > map), not mergemap. Though the inner observable is.. not really an observable, but helpful to use simplier operators when possible

src/ui/connectivityBrowser/connectivityBrowser.component.ts
```diff
@@ -174,7 +173,7 @@ export class ConnectivityBrowserComponent implements AfterViewInit, OnDestroy, A
 
     public ngOnDestroy(): void {
       this.subscriptions.forEach(s => s.unsubscribe())
-      this.defaultColorMap && this.setDefaultMap()
+      // this.defaultColorMap && this.setDefaultMap()
       // this.store$.dispatch({
       //   type: SET_CONNECTIVITY_VISIBLE,
       //   payload: false,
```
xgui3783, line 179: remove instead of comment?

xgui3783 (changes requested):

src/services/state/ngViewerState.store.ts
fsdavid: I could not understand how to you mean to change "mergeMap", because with a simple map we cannot call other actions (We will get an error "Actions must have a type property", maybe fixable but I'm not sure that this is a better approach then) and why is current approach worse?

xgui3783: aww, I see, you are trying to return multiple actions.

I was under the impression that the cleanView$ translates to 1 action

we can safely ignore this comment

Though I suspect, this effect does not need to exist in the first place

PR #644, opened 2020-08-25
Fix switching template while connectivity is on - 639

fsdavid:


xgui3783:
This is ok for a temporary fix. but long term, I think we shouldn't have so many components directly coupled to connectivity. 

There are already `viewerState.connectivityRegion` and `viewerState.overwrittenColorMap`, combined with `effects` we shouldn't need to dispatch connectivity related events everywhere

I think we should merge this now, but during refactor, we shouldn't forget to clean up the code.

PR #648, opened 2020-08-31
RF connectivity component

xgui3783:


xgui3783:
hey @fsdavid , I did some refactoring of the connectivity component, but I am not 100% confident that the component still works in all circumstances.

Can you pull the branch and give some feed back on what doesn't work?

fsdavid:
Sure, I will look at it

fsdavid:
I see 2 major bugs:
1 - connectivity colors does not restore when region deselected from both - sidebar top close icon and from layer widget
2 - if we deselect region and then select any region again, connectivity automatically applies - also if we select and return from other parcellation

xgui3783:
@fsdavid I have fixed the bugs you mentioned... I think

Wouldlove to hear if there are other bugs

fsdavid:
Great, I think it has not more bugs.

PR #664, opened 2020-10-05
Move drag&drop to nehuba container

fsdavid:


xgui3783:
Why is this change required?

fsdavid:
> Why is this change required?

Because it is not possible to make drag&drop feature for plugins and everywhere outside Nehuba viewer.

xgui3783:
Can you elaborate on drag and drop off plugin?

fsdavid:
To import user landmarks, one of the ways is drag&drop. Currently, when the user drags the JSON file on the plugin, in parallel IAV expects to drop the file in it. 
image

PR #668, opened 2020-10-07
small api fixes

fsdavid:


src/ui/nehubaContainer/landmarkUnit/landmarkUnit.component.ts
```diff
@@ -81,8 +81,7 @@ export class LandmarkUnit implements OnChanges {
       const nColor = color?.currentValue
         || this.color
         || (highlight?.currentValue || this.highlight)
-        ? HOVER_COLOR
-        : this.color? this.color : NORMAL_COLOR
+        ? this.color? this.color : HOVER_COLOR : NORMAL_COLOR
```
xgui3783, line 85: this part looks abit weird, can we discuss in detail?

xgui3783: I don't like what is being done in this section. We need to clarify what is being done here. 

I will write unit tests to fix this.

fsdavid: Previous change was made by mistake. "this.color? this.color :" - code should have been added before the colon, before the "HOVER_COLOR".
Yeah, the code looks a little bit wierd because I've tried to fix the coloring bug with minimal changes. instead of "this.color? this.color : HOVER_COLOR" we can simply write "this.color || HOVER_COLOR"

xgui3783: I saw the changes you pushed. I am still in the process of writing unit test for this component. I will push to this branch when done.

xgui3783:
@fsdavid can you check https://github.com/HumanBrainProject/interactive-viewer/blob/92ef53a0245d853dfb78816e72130d29a432b006/src/ui/nehubaContainer/landmarkUnit/landmarkUnit.component.spec.ts and verify that the test cases is as you expect?

fsdavid:
> @fsdavid can you check https://github.com/HumanBrainProject/interactive-viewer/blob/92ef53a0245d853dfb78816e72130d29a432b006/src/ui/nehubaContainer/landmarkUnit/landmarkUnit.component.spec.ts and verify that the test cases is as you expect?

Sure, test cases is as we discussed.

PR #672, opened 2020-10-09
minor visual fixes

fsdavid:


xgui3783:
Which issue does this fix? Can you do a before/after for the visual?

fsdavid:
> Which issue does this fix? Can you do a before/after for the visual?

Sure, you can check it here:

image

xgui3783:
I see

I agree re: info btn

but I disagree with introducing special class `..explore-other-template-item` .

It is probably more generalizable if we make the text container have display: inline-block. The height then should be dynamically adjusted depending on content.

fsdavid:
> I see
> 
> I agree re: info btn
> 
> but I disagree with introducing special class `..explore-other-template-item` .
> 
> It is probably more generalizable if we make the text container have display: inline-block. The height then should be dynamically adjusted depending on content.

I guess it does not work here. I've tried to use inline-block, but it's not working.

xgui3783:
Hmm, you are right, d-inline-block doesn't actually work.

I would still avoid hardcoding height information such as these. It will very quickly unmaintainable. 

If there is a visual issue like this, I believe that the issue lies elsewhere. In this case, I believe it is because there is too much information density associated with each line. 

A list can be selectable, but each selection are usually short and succinct. 

- If a selection involves long description, I believe we should use a card
- otherwise, we could use short name of the template + iconography for left/right hemisphere

fsdavid:
Yes, seems the card works well.

xgui3783:
I will checkout the changes tomorrow.

xgui3783:
@fsdavid I believe the `mat-list` parent dom node for the explore region in other ref space `mat-card`s are no longer necessary?

I removed them (added `cursor-default` class)

Lemme know if you approve

fsdavid:
Great, works and looks nice. 
Do you think that cursor "pointer" is not more natural for buttons?

xgui3783:
hmm, I don't know. 

Perhaps cursor pointer makes more sense? Can you make the change?

fsdavid:
Sorry for the late response, it is updated as well as for dataset lists.

xgui3783:
> it is updated as well as for dataset lists

? can you elaborate?

fsdavid:
> > it is updated as well as for dataset lists
> 
> ? can you elaborate?

There were cursor "text" for regional features but actually, they are clickable buttons.

xgui3783:
> There were cursor "text" for regional features but actually, they are clickable buttons.

That part I get. But what do you mean by `dataset lists`?

fsdavid:
> > There were cursor "text" for regional features but actually, they are clickable buttons.
> 
> That part I get. But what do you mean by `dataset lists`?

I meant a list of regional features.

xgui3783:
Now I am thoroughly confused. As far as I can tell, regional features were not touched at all in your PR?

fsdavid:
The last commit was to change the cursor to a pointer for the region sidebar expansion panel buttons (not concrete to any component).
In the beginning, features were not touched, but as the PR name is "minor visual fixes" and one of the commits is correcting the cursor for one of the sidebar expansion panel buttons, I considered it necessary to correct the cursor not only for a single expansion panel. If you are against, please tell. Just, PR and branch are not connected to some component, they are fixing some general small visual issues and one of them where cursor.

xgui3783:
> The last commit was to change the cursor to a pointer for the region sidebar expansion panel buttons (not concrete to any component).

I don't think this is accurate. `.cursor-pointer` class is added to `mat-card` component inside `ng-template#exploreInOtherTmpl`. It *specifically* only applies to `exploreInOtherTmpl` template. I believe region features were not touched in this PR as of f56c5ce . 

I just did a quick pull and inspection of region features, and did not notice any change from the usual.

Can you let me know how do you think region features had changed? otherwise I think this PR is good.

fsdavid:
I see, for the moment I've moved to vs code, which does not automatically save the files, and forgot to click ctrl+s. Actually, there was only adding a "cursor-pointer" class on the "singleDatasetListView" templates (div element of the title). I still think adding class there will be better but we can leave PR as it is.

PR #677, opened 2020-10-19
Bugfix ref space spatial xform

### LINKED ISSUE #647: [bug] After second change of template coordinate transformation back-end does not work
steps to reproduce:

0/ go to /, select human multilevel, select mni152 template space
1/ change position of template on the viewer 
2/ change template to Colin 27 (after changing template position should be the same)
3/ change template again to mni152

expected behavior:
The position should be the same

actual behavior:
The position is restored to the center

xgui3783:
fixes #647 
supercedes #650

### DIFF
--- a/src/atlasViewer/atlasViewer.apiService.service.ts
+++ b/src/atlasViewer/atlasViewer.apiService.service.ts
@@ -447,7 +447,7 @@ export const overrideNehubaClickFactory = (apiService: AtlasViewerAPIServices, g
   }
 }
 
-export const API_SERVICE_SET_VIEWER_HANDLE_TOKEN = new InjectionToken<(viewerHandle)=>void>('API_SERVICE_SET_VIEWER_HANDLE_TOKEN')
+export const API_SERVICE_SET_VIEWER_HANDLE_TOKEN = new InjectionToken<(viewerHandle) => void>('API_SERVICE_SET_VIEWER_HANDLE_TOKEN')
 
 export const setViewerHandleFactory = (apiService: AtlasViewerAPIServices) => {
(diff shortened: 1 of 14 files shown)


xgui3783:
@fsdavid can you check if this PR fixes #647 ?

fsdavid:
The problem was caused by the URL service. As from the first change, the URL was not updating with new navigation parameters.
Also if I'm not wrong about this bug, it was fixed in "Dev navigation fixes #650" PR. 

In addition, this PR also fixes the bug

xgui3783:
@fsdavid I believe the URL not updating is a result of race condition between initialising nehuba viewer and ngrx state update. 

I cannot consistently get the navigation state to set/not set. It only happens sometimes. 

IIRC, the mess with https://github.com/HumanBrainProject/interactive-viewer/blob/a673bc2
/src/ui/viewerStateController/viewerState.useEffect.ts#L140-L250 is our attempt at combating the race con under time pressure, but I was not 100% happy with the bandaid.

Also, setting navigation as previous navigation is not actually the correct approach. The navigation stored in the state will be the previous navigation state, which is (in most cases) different to that of the state returned by spatial backend. Here is how you can reproduce the error:

0/ checkout dev_navigation_fixes, go to /, select human multi level, change template to mni152
1/ move the navigation to somewhere far away from the center of the brain
2/ change template space to colin, wait until everything loads, note the navigation position
3/ copy the url to another window, wait until everything loads, note the navigation position

This is what I got
Screenshot from 2020-10-20 15-31-38

xgui3783:
I should mention that this PR also has a flaw: if race con fails, the navigation state in url does not get populated.

PR #705, opened 2020-10-29
Dev add new connectivity sources

fsdavid:


src/ui/connectivityBrowser/connectivityBrowser.component.ts
```diff
@@ -106,6 +117,32 @@ export class ConnectivityBrowserComponent implements AfterViewInit, OnDestroy{
...
+    }
+
+    getLoadUrl(): string {
+      if (this.datasetList.length && this.selectedDataset) {
+        const selectedDatasetId = this.datasetList.find(d => d.name === this.selectedDataset).id
+        const url = selectedDatasetId? `${this.connectivityUrl}/${selectedDatasetId}` : null
+        this.connectivityLoadUrl.emit(url)
+        return url
+      }
+    }
```
xgui3783, line 135: this fn gets called in an endless loop, since it is used for an input

you can test this by putting a console.log here, and the console log gets called indefinitely. 

Can you please fix this? otherwise, PR look solid.

xgui3783 (changes requested):

PR #716, opened 2020-10-30
Fix region exploring for Julich brain 2.4

fsdavid:


src/ui/parcellationRegion/region.base.spec.ts
```diff
@@ -49,7 +51,8 @@ const mr1rh = {
 
 const mr0lh = {
   labelIndex: 1,
-  name: 'mr0 - left hemisphere',
+  name: 'mr0',
+  status: 'left hemisphere',
   fullId: {
     kg: {
       kgSchema: 'fzj/mock/pr',
```
xgui3783, line 58: this is actually not what I meant.

the existing tests needs to stay,

but the logic needs to accomodate both the `region name - hemisphere` and `name` `status` scenario

xgui3783 (changes requested):

xgui3783:
I would say, just leave the PR, I can take a look at it. I think I can fix this issue.

More important is, if you can take a look at connectivity components, and see if there are any bugs

src/ui/parcellationRegion/region.base.spec.ts
fsdavid: actually yes, the new approach was only for working for Julich brain 2.4 I will fix it for the old version too

fsdavid:
> I would say, just leave the PR, I can take a look at it. I think I can fix this issue.
> 
> More important is, if you can take a look at connectivity components, and see if there are any bugs

I can look at both. I will fix this issue and I can already found a bug in the connectivity browser.

fsdavid:
Now both versions of julich brain should work

xgui3783:
Thanks. 

I will add unit tests then merge

PR #721, opened 2020-11-02
Fix connectivity coloring

fsdavid:


src/ui/connectivityBrowser/connectivityBrowser.component.ts
```diff
@@ -251,7 +251,7 @@ export class ConnectivityBrowserComponent implements OnInit, AfterViewInit, OnDe
     }
 
     public addNewColorMap() {
-      this.defaultColorMap = new Map(getWindow().interactiveViewer.viewerHandle.getLayersSegmentColourMap())
+      !this.defaultColorMap? this.defaultColorMap = new Map(getWindow().interactiveViewer.viewerHandle.getLayersSegmentColourMap()) : null
```
xgui3783, line 254: can you just use an if statement here? it is IMO significantly more readable than the ternary operator. 

ternary operator IMO is only advantageous in assignment

xgui3783 (changes requested):

src/ui/connectivityBrowser/connectivityBrowser.component.ts
```diff
@@ -251,7 +251,9 @@ export class ConnectivityBrowserComponent implements OnInit, AfterViewInit, OnDe
     }
 
     public addNewColorMap() {
-      !this.defaultColorMap? this.defaultColorMap = new Map(getWindow().interactiveViewer.viewerHandle.getLayersSegmentColourMap()) : null
+      if (this.defaultColorMap) {
+        this.defaultColorMap = new Map(getWindow().interactiveViewer.viewerHandle.getLayersSegmentColourMap())
+      }
```
xgui3783: @fsdavid hmm... this is the exact opposite behaviour?

fsdavid: Yeah, you are true, my mistake. Solved

xgui3783 (approved):

Issue #758, opened 2020-11-27
[Bug] After change connectivity source, if new source does not have data it breaks


fsdavid:
Steps to reproduce:

1. Load Cytoarchitectonic maps - v1.18
2. Select region "Area STS2 (STS) - right hemisphere"
3. Expand the connectivity
4. change connectivity from "1000BRAINS study" to "Averaged_FC_JuBrain_184Regions"

Expected Behavior
Instead of a connectivity diagram, it should return a message that connectivity is not available for the current region

actual Bihevior 
It keeps "1000BRAINS study" diagram on the screen

xgui3783:
perhaps related, as of 6181c38da7065d21d8c5fc1ea8bcafe0d3628667, on init, iav tries to fetch `https://connectivity-query-v1-1-connectivity.apps-dev.hbp.eu/v1.1/studies/1345998a08539dc24bda8817c7d5804558d83b1c`, but returns 500

fsdavid:
The rison is a different one. Fixed in d070569f0c15e431153095519845f7cc11efed4f

xgui3783:
fixed by #776

PR #776, opened 2020-11-30
Fix 758 connectivity source does not have a data

fsdavid:


src/services/state/viewerState/selectors.ts
```diff
@@ -71,6 +71,11 @@ export const viewerStateAllRegionsFlattenedRegionSelector = createSelector(
   }
 )
 
+export const viewerStateOverwrittenColorMap = createSelector(
+  state => state['viewerState'],
+  viewerState => viewerState['overwrittenColorMap']
+)
+
```
xgui3783, line 78: can you add `selector` this this variable name?

we should standardise the names in the future, but for now, all selectors should have selector in their name, same goes with action (and in the future, reducer)

xgui3783 (changes requested):
Just a minor change to variable name.

I have not tested it thoroughly, but I trust that you have.

xgui3783:
@fsdavid are you still interested in this PR?

fsdavid:
> @fsdavid are you still interested in this PR?

@xgui3783 Sure, it contains important bug fixes. Thanks for reminding I will change the name now.

fsdavid:
@xgui3783 Also, I see that there are 2 conflicts, I guess you have updated "export-nehuba"? so we should use:
`"export-nehuba":` `"0.0.4",
"hbp-connectivity-component":` "^0.3.18",`

What about the second conflict, have you removed the parallel connectivity viewer for counting connected areas? There was some issue but, the current PR will fix this too. So we can include the component again.

src/ui/connectivityBrowser/run_test.sh
```diff
@@ -0,0 +1,16 @@
+#! /bin/bash
```
xgui3783, line 1: This is probably not intended to be included, no?

src/ui/connectivityBrowser/test-3.sh
```diff
@@ -0,0 +1,35 @@
+#! /bin/bash
```
xgui3783, line 1: neither is this?

src/ui/connectivityBrowser/test.sh
```diff
@@ -0,0 +1,60 @@
+#! /bin/sh
```
xgui3783, line 1: nor this?

xgui3783 (changes requested):

xgui3783:
re: conflict, I simply move the block inside `mat-accordion` element.

src/ui/connectivityBrowser/test.sh
fsdavid: Sorry for my silly mistakes.

fsdavid:
@xgui3783 Please do not merge the PR yet. I've found one more bug which could be connected to the current branch and I will fix this first.

xgui3783:
@fsdavid is this related to the same or different bug? if it's related to a different bug, it'd rather have two PRs addressing them

fsdavid:
@xgui3783 It is related to the same bug. So to say, the current fix caused these bugs. Now they should have been already fixed.

xgui3783:
i am thoroughly confused... why is `SET_OVERRITEN_COLOR_MAP` needed?

fsdavid:
> SET_OVERRITEN_COLOR_MAP

`SET_OVERRITEN_COLOR_MAP` was used for connectivity visibility but the name was generalized after your comment in the last releases and now it could be used not only for connectivity. 
Why it is used? As I've already mentioned in one of the comments, the expanded connectivity accordion is not directly connected to viewing connectivity on the viewer. Accordion could be expanded but the connectivity view will not be attached in some cases. That's why we have 2 triggers in state, one for the accordion "open/close" state and the second for connectivity colors "visible/invisible" state. I don't now if with this implementation it is possible to fully remove `SET_OVERRITEN_COLOR_MAP` because `ngViewerActionClearView` was confusingly connected to pmaps and accordion "open/close" state.

xgui3783:
@fsdavid I don't know if I follow still. do you have time for a quick zoom?

fsdavid:
@xgui3783 It is connected to #758, as before, the connectivity expansion panel "open/close" state was directly connected with viewing connectivity map on the viewer and now it is not. The bug caused more changes than was expected but it is definitely connected with #748. We could create sub-issues but I think it is not needed and even it could be incorrect. Maybe in the future, if we change the viewer so as we will remove overwritten color maps, it will not be connected to this issue, but now I think it is.

xgui3783:
@fsdavid  it's #758  right?

also, what issue is it causing? #758 is quite clear. From what I can tell, the connectivity map is still connected with open/closing of the panel.

fsdavid:
> @fsdavid it's #758 right?
> 
> also, what issue is it causing? #758 is quite clear. From what I can tell, the connectivity map is still connected with open/closing of the panel.

@xgui3783 Sure it is #758 sorry :). So the case is that at this PR, with the first commit, I've developed more bugs :). The next commits are to fix them, not new ones. 

Yes, the panel's "open/close" state is still connected with the connectivity view (but only if the selected connectivity dataset has one). For the user, working behavior has not changed at all, but now  "open/closing" of the panel now uses "SET_CONNECTIVITY_VISIBLE". As I guess `ngViewerActionClearView`, was created for the same purposes, but for the current implementation I fix the bugs with using both `SET_CONNECTIVITY_VISIBLE` and `ngViewerActionClearView`.

src/services/state/viewerState.store.ts
```diff
@@ -264,7 +264,7 @@ export const ADD_TO_REGIONS_SELECTION_WITH_IDS = `ADD_TO_REGIONS_SELECTION_WITH_
 export const NEHUBA_LAYER_CHANGED = `NEHUBA_LAYER_CHANGED`
 export const SET_CONNECTIVITY_REGION = `SET_CONNECTIVITY_REGION`
 export const CLEAR_CONNECTIVITY_REGION = `CLEAR_CONNECTIVITY_REGION`
-export const SET_CONNECTIVITY_VISIBLE = `SET_CONNECTIVITY_VISIBLE`
+export const SET_OVERRITEN_COLOR_MAP = `SET_OVERRITEN_COLOR_MAP`
```
xgui3783, line 267: can you corrrect typo? `SET_OVERWRITTEN_COLOR_MAP`

xgui3783 (changes requested):
@fsdavid thanks, just 1 final small thing re typo

src/services/state/viewerState.store.ts
fsdavid: Done

xgui3783:
Thanks!

PR #839, opened 2021-01-17
Generate neuron lines VTK string

fsdavid:


src/util/worker.js
```diff
@@ -292,6 +292,147 @@ const processParcRegionAttr = (payload) => {
...
+      `3 ${1 + 8*i} ${0 + 8*i} ${4 + 8*i}\n` +
+      `3 ${1 + 8*i} ${4 + 8*i} ${5 + 8*i}\n`
+  }
+  return returnString
+}
+
+const getPerpendicularPointsForLine = (A, B, scale) => {
+
+  const wi = scale * 230000
+  const he = scale * 230000
```
xgui3783, line 377: How/why was the value 2.3e6 used here? was it arbitrary? I think a more rounded number (1e7) would make much more sense.

src/util/worker.js
```diff
@@ -292,6 +292,147 @@ const processParcRegionAttr = (payload) => {
...
+    })
+    return returnString
+  }
+
+  const coordinateLength = lineCoordinates.length
+
+  const vtk = '# vtk DataFile Version 2.0\n' +
+    'Created by worker thread at https://github.com/HumanBrainProject/interactive-viewer\n' +
+    'ASCII\n' +
+    'DATASET POLYDATA\n' +
```
xgui3783, line 332: use `vtkHeader` rather than repeat the string (also easier in the future, if we would like to change the header)

src/util/worker.js
```diff
@@ -292,6 +292,147 @@ const processParcRegionAttr = (payload) => {
...
+    }
+  }
+
+  const lineCoordinatesArrayToString = () => {
+    let returnString = ''
+    lineCoordinates.forEach(lc => {
+      returnString += getPerpendicularPointsForLine(lc[0], lc[1], scale)
+    })
+    return returnString
+  }
```
xgui3783, line 325: why not just use a IIFE here, rather than assigning a function, then invoking it later?

src/util/worker.js
```diff
@@ -292,6 +292,147 @@ const processParcRegionAttr = (payload) => {
...
+const getColorIds = (n) => {
+  let returnString = ''
+  for (let i=0; i<n-1; i++){
+    returnString += '0\n'
+  }
+  returnString += '0'
+  return returnString
+}
+
+const getPolygonStringByNumber = (n) => {
```
xgui3783, line 354: it's better to be a bit more descriptive, for code maintainability later on. What does n stand for?

src/util/worker.js
```diff
@@ -292,6 +292,147 @@ const processParcRegionAttr = (payload) => {
...
+        data.x[i-1] * plotyMultiple,
+        data.y[i-1] * plotyMultiple,
+        data.z[i-1] * plotyMultiple,
+      ], [
+        data.x[i] * plotyMultiple,
+        data.y[i] * plotyMultiple,
+        data.z[i] * plotyMultiple,
+      ]])
+    }
+  }
```
xgui3783, line 317: this means you are not rendering point data at all? this could be an issue, but we can look at it in a future patch

src/util/worker.js
```diff
@@ -292,6 +292,147 @@ const processParcRegionAttr = (payload) => {
...
+      `3 ${4 + 8*i} ${6 + 8*i} ${7 + 8*i}\n` +
+      `3 ${2 + 8*i} ${6 + 8*i} ${7 + 8*i}\n` +
+      `3 ${2 + 8*i} ${3 + 8*i} ${7 + 8*i}\n` +
+      `3 ${3 + 8*i} ${1 + 8*i} ${7 + 8*i}\n` +
+      `3 ${1 + 8*i} ${5 + 8*i} ${7 + 8*i}\n` +
+      `3 ${2 + 8*i} ${0 + 8*i} ${6 + 8*i}\n` +
+      `3 ${0 + 8*i} ${4 + 8*i} ${6 + 8*i}\n` +
+      `3 ${1 + 8*i} ${0 + 8*i} ${4 + 8*i}\n` +
+      `3 ${1 + 8*i} ${4 + 8*i} ${5 + 8*i}\n`
+  }
```
xgui3783, line 370: I think use a single string literal, rather than lots of `\n` and `+`

Also, I think you can take the `8*i` out of the loop, would improve the perf just a tiny little bit.

src/util/worker.js
```diff
@@ -292,6 +292,147 @@ const processParcRegionAttr = (payload) => {
...
+  return returnString
+}
+
+const getPerpendicularPointsForLine = (A, B, scale) => {
+
+  const wi = scale * 230000
+  const he = scale * 230000
+
+  let u = A.map((item, index) => {
```
xgui3783, line 379: re above, ideally, we can have more readable variable name for maintainability.

src/util/worker.js
```diff
@@ -292,6 +292,147 @@ const processParcRegionAttr = (payload) => {
...
+  const b = sumArrays(A, RMul([-wi,he,0]))
+  const c = sumArrays(A, RMul([wi,-he,0]))
+  const d = sumArrays(A, RMul([-wi,-he,0]))
+
+  const e = sumArrays(B, RMul([wi,he,0]))
+  const f = sumArrays(B, RMul([-wi,he,0]))
+  const g = sumArrays(B, RMul([wi,-he,0]))
+  const h = sumArrays(B, RMul([-wi,-he,0]))
+
+  return `${a.join(' ')}\n ${b.join(' ')}\n ${c.join(' ')}\n ${d.join(' ')}\n ${e.join(' ')}\n ${f.join(' ')}\n ${g.join(' ')}\n ${h.join(' ')}\n `
```
xgui3783, line 432: it had not occurred to me that the maths involved is actually quite involved. 

`export_nehuba` actually exposes some glMatrix, which should have made generating the cuboids much simpler and much easier to read. 

The problem is that `export_nehuba` is a global object on the main thread and worker threads do not get access to the main thread. 

The other solution is to import another copy of glmatrix purely for the calculation.

hmm, I don't know the best approach fo this one yet. I will have to have a look.

xgui3783 (changes requested):
Thanks for the PR.

There are a few minor things

I will take a look at the calculation of the cuboid vertices.

src/util/worker.js
fsdavid: Sorry did not understand which part do you think is hard to read or generates cuboids difficult? 
Does `export_nehuba` find perpendicular points around the line and creates the cuboid with them?
So do you think that we need to change the full approach to using cuboids with `export_nehuba`?

src/util/worker.js
fsdavid: It is 2.3e5 used. 2.3e6 is too weight as well as 1e7 and 1e6. We can use 2e5 if it makes sense.

src/util/worker.js
fsdavid: Cannot take `i` out of the loop. Just will multiply it single time.

`+` is here to make code more readable, do you think it affects performance?

fsdavid:
Thanks for the comments. Changes were made. Please have a look.

src/util/worker.js
xgui3783: Sorry, I didn't mean out of the loop, but before the return string, so 8*i is calculated once, rather than 24 times

xgui3783: re: string literal vs `+` I did some quick tests, and it seems the performance is very similar. we leave the string literal be, I guess.

src/util/worker.js
xgui3783: I mean we should use a number that makes physical sense. 1e6 would translate to: 

> scale of 1 === 1 mm width and height 

2.3e5 would translate to

> scale of 1 === 0.23mm, which does not make a lot of sense

src/util/worker.js
```diff
@@ -292,6 +292,144 @@ const processParcRegionAttr = (payload) => {
...
+
+  const vtk = `${vtkHeader}\n` +
+    `POINTS ${coordinateLength*8} float\n` +
+    (() => {
+      let returnString = ''
+      lineCoordinates.forEach(lc => {
+        returnString += getPerpendicularPointsForLine(lc[0], lc[1], scale)
+      })
+      return returnString
+    })() +
```
xgui3783, line 329: I actually mean assign IIFE a value, then substitute the value here, rather than assign the the value as a function, and invoke the function here. 

Personally, I like the original approach (but invoke the function, rather than invoke it later)

src/util/worker.js
```diff
@@ -351,18 +353,18 @@ const getLineDataVtkPolygonStringWithNumber = (neuronCoordinateLength) => {
...
+       3 ${4 + neuronNumber} ${5 + neuronNumber} ${7 + neuronNumber}
+       3 ${4 + neuronNumber} ${6 + neuronNumber} ${7 + neuronNumber}
+       3 ${2 + neuronNumber} ${6 + neuronNumber} ${7 + neuronNumber}
+       3 ${2 + neuronNumber} ${3 + neuronNumber} ${7 + neuronNumber}
+       3 ${3 + neuronNumber} ${1 + neuronNumber} ${7 + neuronNumber}
+       3 ${1 + neuronNumber} ${5 + neuronNumber} ${7 + neuronNumber}
+       3 ${2 + neuronNumber} ${0 + neuronNumber} ${6 + neuronNumber}
+       3 ${0 + neuronNumber} ${4 + neuronNumber} ${6 + neuronNumber}
+       3 ${1 + neuronNumber} ${0 + neuronNumber} ${4 + neuronNumber}
+       3 ${1 + neuronNumber} ${4 + neuronNumber} ${5 + neuronNumber}\n`
```
xgui3783: Sorry, I think I was not very clear. I did some test, and found that using one string literal (as you did here) vs concat string with `+` did not seem very different in terms of performance. I concluded my comment by saying

> let's leave it as string literal

what I really should have said is

> leave it as concat string with `+`

The unfortunate thing is the current code will not work, because of white spaces. 

Can you undo this part?

The above `lineCoordinatesArrayToString` is exactly what I meant.

xgui3783 (changes requested):

src/util/worker.js
```diff
@@ -292,6 +292,144 @@ const processParcRegionAttr = (payload) => {
...
+      `3 ${0 + neuronNumber} ${4 + neuronNumber} ${6 + neuronNumber}\n` +
+      `3 ${1 + neuronNumber} ${0 + neuronNumber} ${4 + neuronNumber}\n` +
+      `3 ${1 + neuronNumber} ${4 + neuronNumber} ${5 + neuronNumber}\n`
+  }
+  return returnString
+}
+
+const getPerpendicularPointsForLine = (A, B, scale) => {
+
+  const lineWeight = 2e5
```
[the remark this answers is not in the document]
fsdavid: I would like too. but I think for the time being the conditional number should be ok.

xgui3783 (approved):

Issue #844, opened 2021-01-20
[bug] getUserToSelectRegion is not properly intercepted


xgui3783:
as of 75a99ae30d8e85df924b58e11a8fe59abd236057

to reproduce:

0/ go to /, select human multilevel
1/ start jugex plugin, scan for region 1
2/ click any region (e.g. hoc1)

expected behaviour: hoc1 selected by webjugex, hoc1 not selected by viewer

actual beahviour: hoc1 selected by webjugex, hoc1 selected by viewer

xgui3783:
fixed by #849

PR #847, opened 2021-01-27
Fix plugin region selection #844

fsdavid:


xgui3783:
Thanks for the PR, and I can see how you attempted to fix the issue. 

However, I think this is not how I would like to handle user click viewer event. 

`NehubaGlueCmp` should not need to know which services requested access for viewer click. 

I would like to use the middleware like design, where each interceptor can be privy to the event, then can choose to propagate the event, or not. 

During the refactor, I have incorrectly implemented the interceptor service, and I am just now working on the fix (& test to prevent regression like this in the future)

fsdavid:
Thanks. I'm closing the PR.

Issue #889, opened 2021-02-26
[FR] migrate to use brainscapes-api instead of existing backend


xgui3783:
endpoints to be migrated:

- [x] atlas (https://github.com/HumanBrainProject/interactive-viewer/issues/908)
- [x] template (https://github.com/HumanBrainProject/interactive-viewer/issues/908)
- [x] parcellation (https://github.com/HumanBrainProject/interactive-viewer/issues/908)
- [x] region (https://github.com/HumanBrainProject/interactive-viewer/issues/908)
- [x] fetch regional datasets (dependency: https://github.com/FZJ-INM1-BDA/siibra-python/issues/21)
- [x] regional prob map (one's provided in siibra-python does not work, because 1/ cors, 2/ incorrect xyzt_header 3/ datatype) (018a084c0144acc0657430643186749cc844a122)
- [x] connectivity (https://github.com/HumanBrainProject/interactive-viewer/issues/909)
- [ ] #976
- [x] detailed regional features (receptor density) (https://github.com/HumanBrainProject/interactive-viewer/issues/910)
- [ ] #975

xgui3783:
perhaps we should break this card into smaller cards?

- one for altas/template/parcellation/regoin
- one for connectivity
- one for region in different template
- one for regional features
?

fsdavid:
Yes. i agree. I will do that

xgui3783:
@fsdavid I have some time, I can also help. if/when you create the issue, link it to this card under the correct category, eh?

xgui3783:
@fsdavid I have created issues as discussed

I merged `region diff tmpl` with `atlas / tmpl / parc / region`, seeing that they are very much connected

can you assign youself the card(s) you are currently working on?

fsdavid:
Currently, I've started working on `atlas / tmpl / parc / region` would like to change the approach a little bit (load them after selection but I'm not sure yet what problems I will meet.

PR #902, opened 2021-03-04
Fix connectivity 898

fsdavid:


e2e/src/selecting/region.prod.e2e-spec.js
```diff
@@ -74,5 +74,25 @@ describe('> selecting regions', () => {
...
+
+      await newPage.cursorMoveToAndClick({
+        position: [width * 3 / 4, height / 4]
+      })
+      await newPage.wait(500)
+      await newPage.waitForAsync()
+      const elementExists = await newPage.elementExists(`connectivity-browser`)
+      expect(elementExists).toBeTrue()
+
+    })
```
xgui3783, line 96: this test should actually explicitly be false. ie on page load, connectivity-browser should **not** be init. - because it is set to load lazily.

fsdavid:
From the last implementation, there were two connectivity components 1 for the connected numbers which are located on the expansion panel header, and the second one for connectivity itself (as an angular component for connectivity was moved to its own module), which was not the correct approach I guess. As the connectivity component is an external web component we need to run it anyways and get the number, so instead of run 1 connectivity web component on the nehuba container and second on the angular "component" itself we have expanding panel to get the connection number first and then closing the panel and use half resources. Maybe in the future when we will fully integrate Brainscapes API, if we will get connection number for the region without fetching full connectivity we could be able to change current approach but for now, I think it is the most efficient and performant way.

e2e/src/selecting/region.prod.e2e-spec.js
fsdavid: It should not because we need the connection number and having a connectivity web component two times in a document does not look like a better approach.

fsdavid:
> I suspect this has something to do with getting connectivity data? I suspect if you implement #889 for connectivity data, this issue may be resolved automatically?

Yes, maybe after Brainscapes integration the ability to use lazy loading will be more realistic.

fsdavid:
The third approach is that we can change the connectivity web component so that it will be able to be used only to get the number of connections and not the diagram but I do not know if it is the best approach. Maybe for the time being the current one is the best?

e2e/src/selecting/region.prod.e2e-spec.js
xgui3783: more reason to have angular take care of fetching of connectivity metadata + data, and leave the web component as dumb as possible: take data input, and render bar graph. maybe dispatch click / hover events, nothing more.

xgui3783:
> As the connectivity component is an external web component we need to run it anyways and get the number,

at the time, when brainscape didn't exist, this might have been the approach we went for. But with brainscape unifityin the api into a single end point, it is far better to make the component dumb, and let angular do the fetching.

fsdavid:
> > As the connectivity component is an external web component we need to run it anyways and get the number,
> 
> at the time, when brainscape didn't exist, this might have been the approach we went for. But with brainscape unifityin the api into a single end point, it is far better to make the component dumb, and let angular do the fetching.

I agree. So if we need to have a fix ASAP "before we integrate Brainscapes" we can merge current PR (and I will add "ToDo" comments to remove it after integration). If we can wait some time I will close the PR.

xgui3783:
you are working to integrate brainscapes api any way, right?

e2e/src/selecting/region.prod.e2e-spec.js
fsdavid: yeah we could register this as the future feature request.

fsdavid:
> you are working to integrate brainscapes api any way, right?

Yes. Currently, I'm still on quick view but from tomorrow I hope I will back to it.

fsdavid:
Fixed with: https://github.com/FZJ-INM1-BDA/siibra-explorer/pull/949

PR #920, opened 2021-03-22
Add quick tour

fsdavid:


src/ui/quickTour/quickTour.directive.ts
```diff
@@ -0,0 +1,68 @@
...
+    onClick(){
+
+      this.overlayRef = this.overlay.create({
+
+        height: '0px',
+        width: '0px',
+
+        hasBackdrop: false,
+        positionStrategy: this.overlay.position().global(),
+        // panelClass: ['w-100', 'h-100'],
```
xgui3783, line 34: please remove

src/ui/quickTour/quickToutCmp/quickTour.component.ts
```diff
@@ -0,0 +1,327 @@
...
+  styleUrls : ['./quickTour.style.css',],
+})
+export class QuickTourComponent implements AfterViewInit, OnDestroy {
+
+  @Output() destroy = new EventEmitter()
+  public currentTip = 0
+  public numberOfSteps = null
+  private observers: any[] = []
+
+  public slides: any[] = [
```
xgui3783, line 15: I think this approach is not very maintainable. 

For example, if we start to shuffling elements, we will have to recalculate the positions, and if we want to add new slides, we will have to calculate the position of those.

I think we should calculate these dynamically.

src/ui/quickTour/quickToutCmp/quickTour.component.ts
```diff
@@ -0,0 +1,327 @@
...
+      arrowTransform: 'scaleX(-1)',
+      arrow: 'arrow1',
+
+      step: 9,
+    },
+  ]
+
+
+  constructor() {
+    this.numberOfSteps = new Array([...this.slides].filter(s => s.step).pop()['step'])
```
xgui3783, line 158: why not just use map here?

also, no need for `new Array()` constructor. `filter` and `map` already creates new copies of arrays

src/ui/quickTour/quickToutCmp/quickTour.component.ts
```diff
@@ -0,0 +1,327 @@
...
+
+    const observerConfig = {
+      attributes: true,
+      childList: true,
+      characterData: true
+    }
+
+    observer.observe(layerSelectorEl, observerConfig)
+
+    this.observers.push(observer)
```
xgui3783, line 182: I think I see what you are trying to do here, but I think by maintaining a dynamic list of help icons, this should not be needed

src/ui/quickTour/quickToutCmp/quickTour.component.ts
```diff
@@ -0,0 +1,327 @@
...
+      this.currentTip -= 1
+    }
+
+    // Check sidebar status when on open/close
+    const clickInsideCdkOverlay = document.querySelector('.cdk-overlay-container')?.contains(event.target)
+    if (!clickInsideCdkOverlay) {
+      this.checkSidebarStatus()
+    }
+
+  }
```
xgui3783, line 213: this is way too complicated and way to bespoke, and lacks maintainability.

overlapping mat drawer may be an issue, but I do not think this is the way to solve it.

src/ui/quickTour/quickToutCmp/quickTour.component.ts
```diff
@@ -0,0 +1,327 @@
...
+
+
+  goBack() {
+    // Click document for closing menu
+    if (this.slides[this.currentTip]?.documentClickOnNext) {
+      document.body.click()
+    }
+    // Moeve to prev tip
+    this.currentTip = this.slides.findIndex(s => s.step === this.slides[this.currentTip].step-1)
```
xgui3783, line 222: again, logic like this is prone to error. 

For example, we remove step 3, but step 2 and step 4 are still present.

Rather than keep the steps in property, should use linked list, which is built for use cases like this.

src/ui/quickTour/quickToutCmp/quickTour.component.ts
```diff
@@ -0,0 +1,327 @@
...
+
+    if (this.slides[this.currentTip]?.autoNext === 'layerSelectorOpened' && layerSelectorEl['dataset']['opened'] === 'true') {
+      this.currentTip += 1
+    }
+  }
+
+  private lastTipForSidebar = null
+  private drawerOpened = false
+  private drawerExpanded = false
+  checkSidebarStatus() {
```
xgui3783, line 277: again, this is way too complicated and unmaintainable.

src/ui/quickTour/quickToutCmp/quickTour.style.css
```diff
@@ -0,0 +1,90 @@
...
+
+.stepCircle {
+    width: 10px;
+    height: 10px;
+    border-radius: 5px;
+}
+
+/*.cover {*/
+/*    pointer-events: none; z-index: 900;*/
+/*}*/
```
xgui3783, line 90: remove please

src/ui/quickTour/quickToutCmp/quickTour.temlate.html
```diff
@@ -0,0 +1,137 @@
...
+        <path class="stokeColor" d="M 302.9312615546566 333.07007358944463 C 302.65366244614114 329.72018099572654 297.8531401351693 325.79953737428235 300.1758117675781 323.3697204589844 C 300.7803565677791 322.73728789250043 311.83416748046875 323.5789384146184 311.83416748046875 323.81036376953125" style="fill: rgba(0, 0, 0, 0);"/>
+        <path class="stokeColor" d="M 463.05231850248373 323.3282565340343 C 465.30665142189343 323.3282470703125 468.0848881713475 320.89690392424393 469.7531433105469 322.41314697265625 C 470.9482381993743 323.499344326286 470.57628053827574 331.8448791503906 469.25830078125 331.8448791503906" style="fill: rgba(0, 0, 0, 0);"/>
+    </svg>
```
xgui3783, line 105: hmm, I don't know if it is actually better to use a template or an actual svg for this.

xgui3783 (changes requested):
I think we should rethink the strategy in which we construct the quick tour.

The current method is very unmaintainable. 

Let's take a look at how `mat-menu` <https://material.angular.io/components/menu/overview> implemented the menu. 

It somewhat does exactly what we are doing, but very much extensible, and understandable.

fsdavid:
@xgui3783 Could you please look at the new commit? I've made some major changes after our discussion.

src/ui/quickTour/quickToutComponent/quickTour.component.ts
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    this.tipHidden = false
+  }
+
+  positionChangedByArrow = false
+
+  get isLast() {
+    return this.quickTourService.currentTip$.value === [...this.quickTourService.quickTourThisDirectives].pop()
+  }
```
xgui3783, line 204: need to check if this getter gets called in very render cycle

src/ui/quickTour/quickToutComponent/quickTour.temlate.html
```diff
@@ -0,0 +1,141 @@
...
+                    overwritePos?.arrowAlign === 'left'? 'flex-column align-items-start' :
+                    overwritePos?.arrowAlign === 'center'? 'flex-column align-items-center' :
+                    overwritePos?.arrowAlign === 'right'? 'flex-column align-items-end' :
+                    overwritePos?.arrowPosition.includes('left')? 'align-items-start' : 'align-items-end'
+                :overwritePos?.arrowPosition.includes('bottom')?
+                    overwritePos?.arrowAlign === 'left'? 'flex-column-reverse align-items-end' :
+                    overwritePos?.arrowAlign === 'center'? 'flex-column-reverse align-items-center' :
+                    overwritePos?.arrowAlign === 'right'? 'flex-column-reverse align-items-end' :
+                    overwritePos?.arrowPosition.includes('left')? 'align-items-start' : 'align-items-end'
+                : '']">
```
xgui3783, line 26: this probably should be condensed into a single public property

this is not very readable.

src/ui/quickTour/quickToutComponent/quickTour.temlate.html
```diff
@@ -0,0 +1,141 @@
...
+                    marginTop: this.overwritePos?.arrowMargin?.top && this.overwritePos.arrowMargin.top + 'px',
+                    marginBottom: this.overwritePos?.arrowMargin?.bottom && this.overwritePos.arrowMargin.bottom + 'px'}">
+        <ng-content *ngTemplateOutlet="arrow"></ng-content>
+    </div>
+
+    <div #tipCard class="d-flex align-items-center justify-content-center h-100 tipCard pt-3 pr-3 pl-3"
+         [ngStyle]="{width: slideWidth + 'px'}">
+
+<!--         [ngStyle]="{margin: slides[currentTip]?.textMargin? slides[currentTip].textMargin : null,-->
+<!--                    maxWidth: slides[currentTip]?.textWidth? slides[currentTip].textWidth : null}">-->
```
xgui3783, line 41: remove please

src/ui/quickTour/quickToutComponent/quickTour.temlate.html
```diff
@@ -0,0 +1,141 @@
...
+</div>
+
+
+<ng-template #arrow>
+    <ng-content *ngTemplateOutlet="overwritePos?.arrow === 'arrow2'? arrow2:
+                overwritePos?.arrow === 'arrow3'? arrow3:
+                overwritePos?.arrow === 'arrow4'? arrow4:
+                overwritePos?.arrow === 'arrow5'? arrow5:
+                arrow1"></ng-content>
+</ng-template>
```
xgui3783, line 88: probably ... we should extrapolate this into a separate component

src/ui/topMenu/topMenuCmp/topMenu.components.ts
```diff
@@ -51,6 +51,14 @@ export class TopMenuCmp {
   public pluginTooltipText: string = `Plugins and Tools`
   public screenshotTooltipText: string = 'Take screenshot'
 
+  quickTourData = {
+    description: 'These icons provide access to plugins, pinned datasets, and user documentation. Use the profile icon to login with your EBRAINS account.',
```
xgui3783, line 55: ditto above, extrapolate into common/constants

xgui3783 (changes requested):
overall, much better than first approach.

Can you migrate the desc to a separate file first?

We can probably discuss how to do the overwrite position in a call in the future.

src/ui/quickTour/quickToutComponent/quickTour.component.ts
```diff
@@ -0,0 +1,206 @@
...
+    this.top = null
+    this.leftBeforeHardcodeChange = null
+    this.topBeforeHardcodeChange = null
+    this.description = null
+    this.overwritePos$ = null
+    this.overwritePos = null
+    this.order = null
+    this.targetElWidth = null
+    this.targetElHeight = null
+    this.positionChangedByArrow = false
```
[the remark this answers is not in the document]
fsdavid: There are some elements like `targetElWidth` which is necessary for calculations, otherwise, I've removed some.

src/ui/quickTour/quickToutComponent/quickTour.component.ts
```diff
@@ -0,0 +1,206 @@
...
+              } else if (this.leftBeforeHardcodeChange) {
+                this.left = this.leftBeforeHardcodeChange
+              }
+              if (op.top) {
+                this.topBeforeHardcodeChange = this.top
+                this.top = op.top
+              } else if (this.topBeforeHardcodeChange) {
+                this.top = this.topBeforeHardcodeChange
+              }
+              setTimeout(() => this.calculateArrowPosition())
```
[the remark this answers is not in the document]
fsdavid: Made all the changes, but anyways `settimeout` is necessary for some slides, so I suggest keeping it.

src/ui/quickTour/quickToutComponent/quickTour.component.ts
```diff
@@ -0,0 +1,206 @@
...
+    this.top = y
+    this.description = tip.description
+    this.overwritePos$ = tip.overwritePos$
+    this.order = tip.order
+
+    this.targetElWidth = width
+    this.targetElHeight = height
+    this.currentTip = tip
+
+    if (this.tipCardEl) setTimeout(() =>this.calculatePosition(tip, width, height))
```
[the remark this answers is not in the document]
fsdavid: It is needed because current tip sizes are included in calculations and the tip needs to be added to the document.

src/ui/quickTour/constrants.ts
```diff
@@ -0,0 +1,20 @@
...
+    description: string
+    position?: QuickTourPosition
+}
+
+export interface QuickTourPosition {
+    // Position of tip
+    position?: 'top' | 'top-right' | 'right' | 'bottom-right' | 'bottom' | 'bottom-left' | 'left' | 'top-left'
+    align?: 'right' | 'left' | 'top' | 'bottom' | 'center'
+    left?: number
+    top?: number
```
xgui3783, line 12: why are `align`, `top` and `left` property needed?

src/ui/quickTour/constrants.ts
```diff
@@ -0,0 +1,20 @@
...
+
+    // Position of arrow
+    arrow?: string
```
xgui3783, line 15: this property needs to be more explicity. `arrowType`?

src/ui/quickTour/constrants.ts
```diff
@@ -0,0 +1,20 @@
...
+    arrowPosition?: 'top' | 'top-right' | 'right' | 'bottom-right' | 'bottom' | 'bottom-left' | 'left' | 'top-left'
+    arrowAlign?: 'right' | 'left' | 'top' | 'bottom' | 'center'
+    arrowMargin?: {top?: number, right?: number, bottom?: number, left?: number}
+    arrowTransform?: string
```
xgui3783, line 19: These properties should not need necessary. 

if the `quickTourThis` directive specifies `position` attribute, these properties can and should be calculatable.

src/ui/quickTour/quickTour.service.ts
```diff
@@ -11,14 +11,29 @@ export class QuickTourService {
...
+    if (dir.order > this.quickTourThisDirectives[i].order) {
+      if (!this.quickTourThisDirectives[i+1]) {
+        return this.quickTourThisDirectives.push(dir)
+      } else if (dir.order <= this.quickTourThisDirectives[i+1].order) {
+        return this.quickTourThisDirectives.splice(i+1, 0, dir)
+      } else {
+        this.insertSorted(i+1, dir)
+      }
+    } else {
+      this.quickTourThisDirectives.splice(i, 0, dir)
```
xgui3783, line 31: this is more complicated than it should be. I believe you really just need a `findIndex` call, then followed by a `splice` call

src/ui/quickTour/quickTour.service.ts
```diff
@@ -32,8 +47,14 @@ export class QuickTourService {
...
-  public backSlide() {
+  public previousSlide() {
     this.currSlideNum--
     this.currentTip$.next(this.quickTourThisDirectives[this.currSlideNum])
   }
+
+  changeDetected(order) {
+    if (this.currentTip$.value && order === this.currentTip$.value.order) {
+      this.currentTip$.next(this.quickTourThisDirectives[this.currSlideNum])
+    }
```
xgui3783, line 58: I slightly disagree how this should be implemented. 

Rather than reusing `currentTip$`, I _think_ it should be a separate observable called `forceUpdate$` or `detecteChange$`,  of type `Observable<null>` and the quickTour cmp will subscribe to it. 

On emit, quickTour cmp should recalculate/rerender.

src/ui/quickTour/quickTourThis.directive.ts
```diff
@@ -1,32 +1,20 @@
...
-        arrow?: string
-        recalculate?: boolean
-    }>
+export class QuickTourThis implements OnInit, OnChanges, OnDestroy {
 
     @Input('quick-tour-order') order: number = 0
     @Input('quick-tour-description') description: string = 'No description'
-    @Input('quick-tour-position') position: 'top' | 'top-right' | 'right' | 'bottom-right' | 'bottom' | 'bottom-left' | 'left' | 'top-left'
-    @Input('quick-tour-align') align: 'right' | 'left' | 'top' | 'bottom' | 'center'
+    @Input('quick-tour-position') position: QuickTourPosition
```
xgui3783: I like this. This is much cleaner and easier to understand

src/ui/quickTour/quickToutComponent/quickTour.component.ts
```diff
@@ -19,25 +18,25 @@ import {map, switchMap} from "rxjs/operators";
...
-  slideWidth = 300
+  public slideWidth = 300
+  private tip: any
 
-  private tip
+  public isLast = false
+  public left: number
+  public top: number
+  private topHardcoded = false
+  private leftHardcoded = false
```
xgui3783, line 28: correct me if I am missing something, but these two properties should no longer be needed?

xgui3783 (commented):
I will take a look at the branch today, and may push some PRs to the branch, and we can discuss pros/cons

in the mean time, can you resolve the conflicts with current `dev` branch?

src/ui/quickTour/quickToutComponent/quickTour.component.ts
fsdavid: > correct me if I am missing something, but these two properties should no longer be needed?

I've added these properties in the last commit, they are needed because it is possible to hardcode only "top" or "left" props.

src/ui/quickTour/quickTour.service.ts
fsdavid: `currentTip$` - means tip object which is currently set (not only the index) Why you disagree with updating `currentTip$` when the current tip updated? is there a case that we will not need to upgrade `currentTip$` when it actually updates?

src/ui/quickTour/quickTour.service.ts
fsdavid: Sorry did not understand what do you mean. It could be complicated on the first look, but it is quite easy logic. As I understand we change the last implementation because of the performance and this logic should be fast.

xgui3783: I can say that expect for `this.quickTourThisDirectives.splice(i, 0, dir)`, I did not understand the reset of the logic.

src/ui/quickTour/constrants.ts
fsdavid: The top and left position could be hardcoded but it is not necessary because it could be attached to the element where the directive is assigned, so if the position is `top`, `align` could be `left`, `center` and `right`.

src/ui/quickTour/constrants.ts
fsdavid: arrow could be arount tip object, sometimes tip position is not enough to calculate arrow position

src/ui/quickTour/quickTour.service.ts
xgui3783: `order` is not equivalent to and uuid, and it's dangerous to treat it so.

It's better to check for strict equivalence

PR #927, opened 2021-04-08
Feat: reworked quick tour functionalities

xgui3783:
@fsdavid as promised, I prepared a PR to your quick tour branch.

- reworked the calculation of position of quick tour component + arrow (they are now completely dynamic)
- added resize observer, which should update the position when window is resized
- reworked force recalculate position

Can you take a look if I have missed anything?


xgui3783:
One thing I should note is, I have no idea how arrows differed from one and another, so I had to recreate them.

fsdavid:
As I see while touring full viewer is darker, I'm not sure that it is better than keep it light. Could you check the google maps approach?
image

fsdavid:
The tours progress indicator is missing.

fsdavid:
Does this arrow look nice? 
image

could we change it back?

fsdavid:
Arrow for the slice view tools are missing - I had the same problem and I got the review that it needs to be at the correct position.
image

fsdavid:
Region search slide is missing

fsdavid:
Automatic changes for the layer selector don't work anymore. The same approach is used also for google maps.
image

xgui3783:
> As I see while touring full viewer is darker, I'm not sure that it is better than keep it light. Could you check the google maps approach?

I actually disagree with google map's approach. I find it very hard to find the tour panel, without the overlay

xgui3783:
> The tours progress indicator is missing.

This is true. This can be added relatively easily, but the danger is that it can change dynamically (depending on elements being created/destroyed etc)

xgui3783:
> Region search slide is missing

How did you handle when search bar is not being shown?

xgui3783:
> Automatic changes for the layer selector don't work anymore. The same approach is used also for google maps.

Hmm, there is a slight difference. Google maps elements do not change drastically. They translate, and (maybe) scale a little bit. 

In the case of layer selector, the entire element element transform into a different element.

xgui3783:
> Does this arrow look nice?
> could we change it back?

I don't mind changing it back, but as discussed, I have no idea about the difference between arrow1-arrow7. 

If you feel strongly about, feel free to push to branch with new arrows, but with more descriptive names how to distinguish the arrows.

xgui3783:
> Arrow for the slice view tools are missing - I had the same problem and I got the review that it needs to be at the correct position.

This can be fixed. I think i missed it yes.

fsdavid:
> > As I see while touring full viewer is darker, I'm not sure that it is better than keep it light. Could you check the google maps approach?
> 
> I actually disagree with google map's approach. I find it very hard to find the tour panel, without the overlay

Yeah, maybe it is a disadvantage but, that's why we can change the style of the `tip` - so we will not have problems at all - users will explore features better and find `tip` cards easier.

fsdavid:
> > The tours progress indicator is missing.
> 
> This is true. This can be added relatively easily, but the danger is that it can change dynamically (depending on elements being created/destroyed etc)

Yeah, I think it should not be the problem - it was Timos suggestion to see the progress. You are right and that's why I think it is better to implement a progress indicator with `point` icons, not with numbers, as it was in the last version.

fsdavid:
> > Region search slide is missing
> 
> How did you handle when search bar is not being shown?

When the sidebar was open, in normal cases region search would be there. (I guess we can also watch if there is dataset preview)

fsdavid:
> > Automatic changes for the layer selector don't work anymore. The same approach is used also for google maps.
> 
> Hmm, there is a slight difference. Google maps elements do not change drastically. They translate, and (maybe) scale a little bit.
> 
> In the case of layer selector, the entire element element transform into a different element.

Yeah, that's why it is useful to have overwritten position. Anyways we will stay into the same component I guess.

xgui3783:
> > > As I see while touring full viewer is darker, I'm not sure that it is better than keep it light. Could you check the google maps approach?
> > 
> > 
> > I actually disagree with google map's approach. I find it very hard to find the tour panel, without the overlay
> 
> Yeah, maybe it is a disadvantage but, that's why we can change the style of the `tip` - so we will not have problems at all - users will explore features better and find `tip` cards easier.

removing the darkened overlay is actually very easy, but I still disagree that it should be be clear.

xgui3783:
> > > The tours progress indicator is missing.
> > 
> > 
> > This is true. This can be added relatively easily, but the danger is that it can change dynamically (depending on elements being created/destroyed etc)
> 
> Yeah, I think dynamic changes should not be the problem - it was Timos suggestion to see the progress. You are right and that's why I think it is better to implement a progress indicator with `point` icons, not with numbers, as it was in the last version.

I agree. I will push commit today or next days for this issue.

xgui3783:
> > > Region search slide is missing
> > 
> > 
> > How did you handle when search bar is not being shown?
> 
> When the sidebar was open, in normal cases region search would be there. (I guess we can also watch if there is dataset preview)

hmm, I thought I added it in, but maybe i took it out by accident. 

I will take a look at the region selection and re add the quick tour to it, if I accdentally took it out

fsdavid:
> > Does this arrow look nice?
> > could we change it back?
> 
> I don't mind changing it back, but as discussed, I have no idea about the difference between arrow1-arrow7.
> 
> If you feel strongly about, feel free to push to branch with new arrows, but with more descriptive names how to distinguish the arrows.

Nice I will do that, just I wanted to mention that there is 1 arrow for almost all the slides. I think that for the elements in corner could be better to have little bit different style arrow than others. I will try to make it work.

fsdavid:
@xgui3783 Thank you so much for the PR :) I will make changes for arrow SVGs.

fsdavid:
> > > > As I see while touring full viewer is darker, I'm not sure that it is better than keep it light. Could you check the google maps approach?
> > > 
> > > 
> > > I actually disagree with google map's approach. I find it very hard to find the tour panel, without the overlay
> > 
> > 
> > Yeah, maybe it is a disadvantage but, that's why we can change the style of the `tip` - so we will not have problems at all - users will explore features better and find `tip` cards easier.
> 
> removing the darkened overlay is actually very easy, but I still disagree that it should be be clear.

Yeah, then let's leave it darker.

xgui3783:
> > > Automatic changes for the layer selector don't work anymore. The same approach is used also for google maps.
> > 
> > 
> > Hmm, there is a slight difference. Google maps elements do not change drastically. They translate, and (maybe) scale a little bit.
> > In the case of layer selector, the entire element element transform into a different element.
> 
> Yeah, that's why it is useful to have overwritten position. Anyways we will stay into the same component I guess. You can look at the third slide of google maps.

I see google map's third slide, I think I see what you mean, I will take a look at how it can be done. 

I still very much disagree that it requires hard coding of the positions, that the positions of the tour can, and should be calculated dynamically.

fsdavid:
> > > > Automatic changes for the layer selector don't work anymore. The same approach is used also for google maps.
> > > 
> > > 
> > > Hmm, there is a slight difference. Google maps elements do not change drastically. They translate, and (maybe) scale a little bit.
> > > In the case of layer selector, the entire element element transform into a different element.
> > 
> > 
> > Yeah, that's why it is useful to have overwritten position. Anyways we will stay into the same component I guess. You can look at the third slide of google maps.
> 
> I see google map's third slide, I think I see what you mean, I will take a look at how it can be done.
> 
> I still very much disagree that it requires hard coding of the positions, that the positions of the tour can, and should be calculated dynamically.

Yeah, it would be perfect, if it will work without hardcoded position.

xgui3783:
@fsdavid I pushed a few more changes. Except for arrow styling, the rest of the issue should be solved.

fsdavid:
> @fsdavid I pushed a few more changes. Except for arrow styling, the rest of the issue should be solved.

Now I will make arrow changes here. 
I still see two issues:

1. Ordering does not work correctly
2. For the atlas layer selector expanded - tip should be on the right-side because it is overflowed at the top or bottom positions
Screen Shot 2021-04-12 at 10 28 02 AM

xgui3783:
@fsdavid can you be more specific which tour has the order incorrectly placed? it would help me pin down the issue.

re: atlas layer browser placement issue: I will look into it.

fsdavid:
> @fsdavid can you be more specific which tour has the order incorrectly placed? it would help me pin down the issue.
> 
> re: atlas layer browser placement issue: I will look into it.

They are different at different times - I guess, sorting algorithm does not work at all. The directive has the input variable `order` and slides should be ordered by this.

xgui3783:
@fsdavid I pushed two commits, which should fix the order issue, and partially fix the dialog position issue. 

smartly positioned quick tour will now be restricted inside the viewport (see below)

Screenshot from 2021-04-12 19-41-09

Absolute positioned tour points (e.g. slice views, see below) will look wonky. I will provide a fix tomorrow.

Screenshot from 2021-04-12 19-40-55

xgui3783:
@fsdavid per our discussion, adaptive region search quick tour is now commited

fsdavid:
> @fsdavid per our discussion, adaptive region search quick tour is now commited

Great looks pretty. Thank you very much.

xgui3783:
@fsdavid anything else we should change?

also, since now there are merge conflicts, I am thinking of merging this to dev instead, and close the other PR, rather than resolve the merge conflicts.

fsdavid:
@xgui3783 No problem - I'm resolving conflicts

fsdavid:
@xgui3783 Conflicts are resolved and will be ready after checks.

xgui3783:
@fsdavid and now tests are failing.

Do you have anything against PR exmp_quickTour -> dev instead of exmp_quickTour -> dev_quick_tour -> dev ?

The commit history will still be there, and your work is credited.

fsdavid:
> @fsdavid and now tests are failing.
> 
> Do you have anything against PR exmp_quickTour -> dev instead of exmp_quickTour -> dev_quick_tour -> dev ?
> 
> The commit history will still be there, and your work is credited.

Hah, I just meant to keep the flow, but resolving conflicts from GitHub was not a good idea :). To avoid force changes I've created a new PR "exmp_quickTour_forMerge" directly to the dev 😬. But we can force remove the last commit from here as well, as you wish. Sorry :)

fsdavid:
I make changes here :)

xgui3783:
now that repo is public, it's a good idea to avoid force pushes. 

I will create a new PR, and ref it here.

PR #931, opened 2021-04-13
Feat: add quick tour

### LINKED ISSUE #899: [FR] Have a quick tour to explore the viewer on a first visit.
Create a quick tour to explore the viewer on a first visit.

xgui3783:
supercedes #920 #930 #927 

fixes #899


xgui3783:
@fsdavid final check to see everything looks good?

fsdavid:
One thing found, on some slides arrow is a little bit left - what do you think? I can make a fix
image

fsdavid:
> One thing found, on some slides arrow is a little bit left - what do you think, is it ok? I can make a fix
> image

fsdavid:
@xgui3783 Now the arrow looks better, and I guess it's ready for merge.

xgui3783:
https://github.com/HumanBrainProject/interactive-viewer/pull/931/commits/d7ebeec34510eb177938731fd13b4c278d648213#diff-494acace84b4eef28c7a5be32d62fb2f05e514dc01f4b957b4d6377623dd3a59R4-R5 we probably need viewbox attr here

fsdavid:
> @fsdavid [d7ebeec#diff-494acace84b4eef28c7a5be32d62fb2f05e514dc01f4b957b4d6377623dd3a59R4-R5](https://github.com/HumanBrainProject/interactive-viewer/commit/d7ebeec34510eb177938731fd13b4c278d648213#diff-494acace84b4eef28c7a5be32d62fb2f05e514dc01f4b957b4d6377623dd3a59R4-R5) we probably need viewbox attr here

It works pretty without the attribute and more problematic with it. I just reduced the arrow size to `2.4 rem` and `transform` to `1.2 rem`, so the arrow is in the middle and works well for all the slides. `viewbox` was taking only 1/4 of SVG and was more problematic to center it for different angels, so with it I could not really center arrow for all slides. Is there some specific reason why we should use `viewbox`?

xgui3783:
> `2.4 rem`

You reduced the size. I am fine with that

> `viewbox` was taking only 1/4 of SVG and was more problematic to center it for different angels, so with it I could not really center arrow for all slides. Is there some specific reason why we should use `viewbox`?

viewBox should have nothing to do with the physical viewport of the svg (which is defined by svg width and height attribute). viewBox really defines the size of the svg canvas (not a canvas element). What it really mean is, if one defined viewBox of `0 0 100 100`, and in the svg, you have an instruction `M100 100`, you are moving to the right most, bottom most point. 

without it, it can be difficult to predict how the svg will behave in different browsers/different versions. In the commit above, `M.5 17.819` can literally mean move to 1 unit to the right, 35 unit down, or 2 unit to the right, 70 unit down, depending on the viewbox definition.

I was sure most programs will export the viewbox property on export. which program are you using to export svg?

fsdavid:
Yes, I was using [vecta.io](https://vecta.io/).

xgui3783:
I double checked, as far as vecta.io is concerned, viewBox can be enabled if, when exporting svg, uncheck the `optimize svg` option.

The default export option seems to hard code the width and height, which is probably ok for print, but dubious for the web. 

The workflow to get a a svg with viewBox attr is as follows:

- file -> page setup -> custom -> 100x100 px
- (scale the svg as the size may have changed)
- export -> export as svg
- optimize slider -> optimization mode -> object (you might also want to tick maintain ID ... etc in future projects)

fsdavid:
@xgui3783 Thank you very much, I've added `viewbox`.

PR #945, opened 2021-05-04
exmpt MVP use siibra-api for template/parc/region

xgui3783:
Co-authored-by: Xiao Gui <xgui3783@gmail.com>
Co-authored-by: Daviti Gogshelidze <daviti1@mail.com>


fsdavid:
Region features on sidebar are missing:

- [ ] Regional features
- [ ] Description
- [ ] Receptor distributions (with that, expansion panel is visible, but it is empty)
- [ ] Connectivity
- [ ] Explore in other templates

xgui3783:
can you take a look why connectivity stopped working? @fsdavid I think it's because siibra-python/siibra-api changed, so that regions are now `{region_name} {left|right}`, instead of `{region_name} - {left|right} hemisphere`

xgui3783:
re: explore in other templates, it has never really been implemented before. can try to implement it?

fsdavid:
> can you take a look why connectivity stopped working? @fsdavid I think it's because siibra-python/siibra-api changed, so that regions are now `{region_name} {left|right}`, instead of `{region_name} - {left|right} hemisphere`

Sure, thanks for the info.

fsdavid:
> re: explore in other templates, it has never really been implemented before. can try to implement it? @fsdavid

Actually, as currently we are fetching all the regions at once, the feature should have been working but seems an error caused by the region name change. Anyways I will not start fixing it, but use API for new implementation.

xgui3783:
> but use API for new implementation.

This is indeed what I meant 👍🏼

PR #946, opened 2021-05-04
migrate connectivity to siibra api

fsdavid:


xgui3783:
Hey I see you updated the connectivity component to rely on siibra-api. 

- is #902 obsolete?
- can you merge `chore_migSiibraApi` into your working branch, then, instead of hardcoding the api endpoint, inject it much like https://github.com/FZJ-INM1-BDA/siibra-explorer/blob/04b0115deeb76057876e4673c290fb6c224805b4/src/util/pureConstant.service.ts#L225 ? Once done, can you raise PR to `chore_migSiibraApi`

PR #949, opened 2021-05-07
Dev conn 2 api

fsdavid:


src/viewerModule/viewerCmp/viewerCmp.component.ts
```diff
@@ -104,6 +106,9 @@ export class ViewerCmp implements OnDestroy {
   private subscriptions: Subscription[] = []
   public viewerLoaded: boolean = false
 
+  public hasConnectivity = false
+  public connectivityNumber = 0
```
xgui3783, line 110: can you move the this logic into a directive inside connectivity module, similar to https://github.com/FZJ-INM1-BDA/siibra-explorer/blob/chore_migSiibraApi/src/atlasComponents/regionalFeatures/bsFeatures/receptor/hasReceptor.directive.ts#L20-L28 ?

you don't have to use a service, encapsulated in the directive is fine.

xgui3783 (changes requested):
Just one minor thing. Everything else looks good.

src/viewerModule/viewerCmp/viewerCmp.component.ts
fsdavid: Sure

PR #951, opened 2021-05-18
Dev adding annotation

fsdavid:


xgui3783 (changes requested):

xgui3783:
@fsdavid  Can you convert this PR to draft? when  ready, convert it to PR, and I can go through and give reviews?

fsdavid:
> @fsdavid Can you convert this PR to draft? when ready, convert it to PR, and I can go through and give reviews?

Sure

xgui3783:
feedback at d5ca35a:

- top right
  - icons should still be available (those not available should be disabled)
  - `annotation mode x` feels out of place 
    - use standard component (panel/card)
    - use larger font
    - x should be `mat-icon-button[color="warn"]`
  - should not be pushed when the side nav toggles
- side nav
  - point/line icon is pushed off:
    Screenshot from 2021-05-25 17-25-25
  - when text field is focus -> esc should not collapse side nav
  - cannot clear landmark name (on clear, new one is automatically populated)
  - allow `alt + enter` to defocus (also related, esc should only defocus, rather than collapse side bar)
  - horizontal rule separating `select` | `point` , `line`, `poly line` | `delete` | `exit`
- missing feature
  - export to sands compatible format
  - import sands compatible format
- bug
  - drag on perspective view moves annotation as well as perspective view

fsdavid:
Thanks for your feedback. 
I have some questions:
Could not understand what do you mean - `icons should still be available (those not available should be disabled)`
Could not reproduce - `point/line icon is pushed off`
Could not understand what do you mean: `allow alt + enter to defocus` (I've added `esc` to defocus input fields)

What about `drag on perspective view moves annotation as well as perspective view` - Do you know if it is possible to get or change perspective view orientation?

xgui3783:
> Could not understand what do you mean: allow alt + enter to defocus (I've added esc to defocus input fields)

I mean `ctrl + enter`, fixed in the comment

xgui3783:
> Could not understand what do you mean - icons should still be available (those not available should be disabled)

I am thinking about user login/logout button, user setting, siibra-explorer about page, etc

xgui3783:
> Could not reproduce - point/line icon is pushed off 

can you show what side nav looks like in your local installation?

fsdavid:
> > What about drag on perspective view moves annotation as well as perspective view - Do you know if it is possible to get or change perspective view orientation?
> 
> I believe so, how are you disabling translation of slice view? it should be exactly the same as preventing translation of the perspective view.

I'm getting the current state of the viewer (var name: `this.navState`), when dragging is started and while dragging I'm running: `this.interactiveViewer.viewerHandle.setNavigationLoc(this.navState)`

fsdavid:
> > Could not reproduce - point/line icon is pushed off
> 
> can you show what side nav looks like in your local installation?

image

xgui3783:
can you check the navside bar is behaving similarly for you on 

https://quick-tour-demo.okd-dev.hbp.eu/

?

fsdavid:
> > Could not understand what do you mean - icons should still be available (those not available should be disabled)
> 
> I am thinking about user login/logout button, user setting, siibra-explorer about page, etc

Yeah, if you prefer we can show and disable them - if it will not be confusing for the users (users could think that app got stuck)

xgui3783:
> Do you mean for description? or for all the fields? what about use ctrl + enter for description and just enter for input fields?

all fields, per trello behaviour

xgui3783:
> > can you check the navside bar is behaving similarly for you on
> > https://quick-tour-demo.okd-dev.hbp.eu/
> > ?
> 
> Yeah sure:
> image

OK my browser renders it differently

We need to inspect it properly.

xgui3783:
i just checked FF, it's even worse:

Screenshot from 2021-05-28 13-40-37

fsdavid:
> i just checked FF, it's even worse:
> 
> Screenshot from 2021-05-28 13-40-37

It works well on my firefox. Could you please check what is hovering icons?

xgui3783:
> Could you please check what is hovering icons?

can you elaborate how I can look for it?

fsdavid:
> > Could you please check what is hovering icons?
> 
> can you elaborate how I can look for it?

Just with selecting element which is hovering icons from dev tools 
image

xgui3783:
I mean, from chrome element inspector, the layout is completely busted

Screenshot from 2021-05-28 13-54-59

fsdavid:
> I mean, from chrome element inspector, the layout is completely busted
> 
> Screenshot from 2021-05-28 13-54-59

Thanks, I will try to make a different approach

src/atlasViewer/atlasViewer.apiService.service.ts
```diff
@@ -83,7 +87,16 @@ export class AtlasViewerAPIServices implements OnDestroy{
...
+  private onMouseClick(ev: any): boolean{
+
+    // If annotation mode is on, avoid region selection
+    let viewerMode
+    this.store.pipe(
+      select(viewerStateViewerModeSelector),
+      take(1)
+    ).subscribe(vm => viewerMode = vm)
+    if (viewerMode === ARIA_LABELS.VIEWER_MODE_ANNOTATING) return false
+
```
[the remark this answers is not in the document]
fsdavid: can you elaborate?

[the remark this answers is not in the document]
xgui3783: apiService should only be used for api calls, from plugins etc. 

for something internal, you can inject clickInterceptor yourself, and intercept mouse clicks

see https://github.com/FZJ-INM1-BDA/siibra-explorer/blob/2992d9959ba9a0c9f75139e5fb006740f01a4d47/src/atlasViewer/atlasViewer.apiService.service.ts#L160-L165 for implementation detail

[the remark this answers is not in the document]
fsdavid: Thanks for explanation

src/atlasComponents/userAnnotations/annotationList/annotationList.component.ts
```diff
@@ -0,0 +1,117 @@
...
+    )
+  }
+
+  saveAnnotation(annotation, singlePolygon = false) {
+    if (annotation.type !== 'polygon' || singlePolygon) {
+      annotation.position1 = annotation.position1.replace(/\s/g, '')
+      annotation.position2 = annotation.position2 && annotation.position2.replace(/\s/g, '')
+      if (annotation.position1.split(',').length !== 3 || !annotation.position1.split(',').every(e => !!e)
+          || ((annotation.position2
+              && annotation.position2.split(',').length !== 3) || !annotation.position1.split(',').every(e => !!e))) {
```
xgui3783, line 66: this part is not very clear at all

it'd be a lot clearer if it is broken into four separate if statement and return if conditions are unmet.

src/atlasComponents/userAnnotations/annotationService.service.ts
```diff
@@ -0,0 +1,460 @@
...
+  "name": CONST.USER_ANNOTATION_LAYER_NAME,
+  "annotationColor": "#ffee00",
+  "annotations": [],
+}
+
+@Injectable()
+export class AnnotationService {
+
+    // Annotations to display on viewer
+    public pureAnnotationsForViewer = []
```
xgui3783, line 21: typing please.

the variable name needs to be clearer also, IMO.

xgui3783 (changes requested):
in general, there are still a lot of issues with how the code is structured. 

in particular annotationService and annotationMode component.

src/atlasComponents/userAnnotations/annotationList/annotationList.style.css
```diff
@@ -0,0 +1,32 @@
+.annotation-content {
+    max-width: 300px;
+    min-width: 300px;
+}
```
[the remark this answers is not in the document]
fsdavid: Unfortunately automatic width changes after adding the first annotation and toolbar hides. What do you think could be done here?

src/atlasComponents/userAnnotations/annotationList/annotationList.style.css
```diff
@@ -0,0 +1,32 @@
...
+
+input, textarea {
+  background: none;
+  border: none;
+}
+
+input:focus, textarea:focus {
+  border-bottom: 2px solid;
+}
```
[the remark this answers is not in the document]
fsdavid: Yes, angular material input has some advantages but is very not in place in this situation, so I recommend keeping the current approach.

src/atlasComponents/userAnnotations/annotationList/annotationList.style.css
```diff
@@ -0,0 +1,32 @@
...
+
+.annotation-name-input {
+  width: 180px;
+}
```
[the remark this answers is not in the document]
fsdavid: it is related to https://github.com/FZJ-INM1-BDA/siibra-explorer/pull/951#discussion_r642842717

src/atlasComponents/userAnnotations/annotationList/annotationList.style.css
```diff
@@ -0,0 +1,32 @@
...
+
+:host-context([darktheme="true"]) .hovering-header {
+    background-color: #737373;
+}
+:host-context([darktheme="false"]) .hovering-header {
+    background-color: rgb(245, 245, 245);
+}
```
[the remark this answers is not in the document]
fsdavid: you mean auto-expand annotation into the list when the user hovers it on the viewer? would not it too much automatic changes of UI?

src/atlasComponents/userAnnotations/annotationService.service.ts
```diff
@@ -0,0 +1,460 @@
...
+        if (foundIndex >= 0) {
+          this.groupedAnnotations[foundIndex] = tr
+        } else {
+          this.groupedAnnotations.push(tr)
+        }
+        this.refreshAnnotationFilter()
+      })
+    }
+
+    refreshAnnotationFilter(filter = null) {
```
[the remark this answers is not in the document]
fsdavid: filtering annotations is not connected only to what users see but also is used for exporting. So to say it is finalized list. Also, I think it is more performant to avoid multiple filters

src/atlasComponents/userAnnotations/annotationList/annotationList.style.css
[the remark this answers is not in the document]
xgui3783: like every other mat-drawer `class="box-shadow-none border-0 pe-none bg-none col-10 col-sm-10 col-md-5 col-lg-4 col-xl-3 col-xxl-2"`

src/atlasComponents/userAnnotations/annotationList/annotationList.style.css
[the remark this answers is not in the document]
xgui3783: I think I know what you mean by 

> is not in place

It is how trello, for eg. deal with editing. I feel somewhat hesitant implementing trello editing style, for the following reasons:

- in trello, single click on column title has a single function -> edit
- in siibra-explorer, single click on column title, user may wish to expand the panel, or edit the name. It is ambiguous. 

I think a much natural expectation is user are able to edit name, desc, position in expanded panel, but the only functionality of the header to expand or collapse the body fo the panel

src/atlasComponents/userAnnotations/annotationList/annotationList.style.css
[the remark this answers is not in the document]
xgui3783: see response above

this should use relative sizing, or flexbox

src/atlasComponents/userAnnotations/annotationList/annotationList.style.css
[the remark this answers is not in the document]
xgui3783: I think I misunderstood this css.

I think we can add a `[lighter-bg]` in `theme.scss` to cover similar use cases.

there is already a `[darker-bg]`

alternatively, we could use `elevation` or `color=primary` to highlight the hovered annotation

src/atlasComponents/userAnnotations/annotationService.service.ts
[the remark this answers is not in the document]
xgui3783: Hmm, it is still quite unclear what the best way to approach this is. 

We can perhaps leave this issue for now. Once we clear up other issues, perhaps we can revisit this issue.

src/atlasComponents/userAnnotations/annotationList/annotationList.template.html
```diff
@@ -0,0 +1,173 @@
...
+                </small>
+            </div>
+        </div>
+    </div>
+    <mat-divider class="mt-2 mb-2"></mat-divider>
+
+    <mat-accordion aria-label="user annotations list"
+                 class="h-100 d-flex flex-column overflow-auto">
+        <mat-expansion-panel #expansion hideToggle *ngFor="let annotation of ans.finalAnnotationList; let i = index; trackBy: identifier">
+            <mat-expansion-panel-header [ngClass]="ans.hoverAnnotation?.id.includes(annotation.id) && 'hovering-header'">
```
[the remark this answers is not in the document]
fsdavid: could you please elaborate?

src/atlasComponents/userAnnotations/annotationList/annotationList.style.css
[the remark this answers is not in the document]
fsdavid: Yeah, I understand, column header in Trello or apps like that are also used for dragging, but I think the current approach looks better. In my opinion editing name or coordinates are not key features as just exploring data, that's why I've avoided the standard input components. UI would too be crowded and looking unnatural if there with standard inputs.

src/atlasComponents/userAnnotations/annotationList/annotationList.template.html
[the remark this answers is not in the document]
xgui3783: I think when I commented, I imagined each annotation can have its own component to separate the concerns.

probably also allow the annotation to be modelled more closely

src/atlasComponents/userAnnotations/annotationList/annotationList.style.css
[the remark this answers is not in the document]
xgui3783: The concern I have is, right now, the behaviour is confusing. 

when one clicks the annotation name, it sometimes toggle the expansion panel, it sometimes doesn't. I feel this is an anti pattern.

To me, it feels more right to remove the trello like interactions altogether, and just have material inputs/textarea in expanded panel.

src/atlasComponents/userAnnotations/annotationList/annotationList.component.ts
```diff
@@ -45,16 +47,19 @@ export class AnnotationList {
...
-    position = position.split(',').map(p => +p * 1e6)
-    this.store$.dispatch(
-      viewerStateChangeNavigation({
-        navigation: {
-          position,
-          positionReal: true
-        },
-      })
-    )
+  navigate(positionString: string) {
```
xgui3783, line 50: this shouldn't be a string, it should be `[number, number, number]`

src/atlasComponents/userAnnotations/directives/keyListener.directive.ts
```diff
@@ -0,0 +1,18 @@
...
+
+  @HostListener('keydown', ['$event'])
+  onKeyDown(e) {
+    if ((e.ctrlKey || e.metaKey) && e.key === 'Enter' || e === 'Escape' || e.key === 'Enter') {
+      e.stopPropagation()
+      this.elementRef.nativeElement.blur()
+    }
+  }
+
+}
```
xgui3783, line 18: I think we can reuse the `key-down-lsitener` directive.

the functionality is very similar

src/atlasComponents/userAnnotations/directives/annotationSwitch.directive.ts
```diff
@@ -0,0 +1,25 @@
...
+import {Store} from "@ngrx/store";
+import {IavRootStoreInterface} from "src/services/stateStore.service";
+
+@Directive({
+  selector: '[annotation-switch]'
+})
+export class AnnotationSwitch {
+
+
+  constructor(private store$: Store<IavRootStoreInterface>) {
```
xgui3783, line 13: the `IavRootStoreInterface` typing is ... very outdated.

we should update the typing, but for now, it's safer to type it to `any`

src/atlasComponents/userAnnotations/directives/annotationSwitch.directive.ts
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  @HostListener('click')
+  onClick() {
+    this.setAnnotatingMode()
+  }
+
+  setAnnotatingMode() {
+    this.store$.dispatch(viewerStateSetViewerMode({payload: ARIA_LABELS.VIEWER_MODE_ANNOTATING}))
+  }
```
xgui3783, line 24: either set this is private, or move into `onClick`

common/constants.js
```diff
@@ -98,7 +98,7 @@
 
     // Annotations
     USER_ANNOTATION_LAYER_NAME: 'user_annotations',
-    USER_ANNOTATION_STORE_KEY: `user_landmarks_demo_1`
+    USER_ANNOTATION_STORE_KEY: `user_landmarks_demo_2`
```
xgui3783: hmm... I don't believe these belong here?

they should be in `component.ts`

If i told you otherwise, shows how much I know =p

src/atlasComponents/userAnnotations/annotationList/annotationList.component.ts
```diff
@@ -64,15 +65,15 @@ export class AnnotationList {
 
   saveAnnotation(annotation, singlePolygon = false) {
```
xgui3783, line 66: typing for annotation?

I don't know if `annotation.position1` is a `string` or `number[]`

fsdavid: Here it could be both. We convert to `number[]` if it is a string. That is one of the reasons why we have `saveAnnotation` method inside `annotationList` component and are not using the service method directly.

src/atlasComponents/userAnnotations/directives/keyListener.directive.ts
fsdavid: I'm not sure which listener you mean. `iav-key-listener` does not work samely

xgui3783: not yet, clearly. But with some modifications, they should work similarly

src/atlasComponents/userAnnotations/annotationList/annotationList.component.ts
xgui3783: This is very confusing, and should be avoided.

I am currently working on an experimental feature that breaks the annotation service into the individual components.

Hopefully, this issue should also be fixed.

fsdavid: Individual components are great, but that should not be very confusing anyways. The user modifies coordinates as a string and it needs to be converted to a number array.

src/atlasComponents/userAnnotations/directives/keyListener.directive.ts
fsdavid: `not yet, clearly.` - I suggest keeping the current directive for the time being.

src/atlasComponents/userAnnotations/annotationList/annotationList.component.ts
xgui3783: it should be transformed to `number[]` before passing to this function. 

functions should be as simple as possible, doing one thing and one thing only. this would make testing and debugging significantly easier.

src/atlasComponents/userAnnotations/directives/keyListener.directive.ts
xgui3783: adding code is easy, removing code is difficult. 

It makes no sense to end up have 2 directives doing 90% of the same things.

src/atlasComponents/userAnnotations/annotationList/annotationList.component.ts
fsdavid: you are right. Here the function name is most confusing. I hope we will move that function to an individual component.

src/atlasComponents/userAnnotations/directives/keyListener.directive.ts
fsdavid: Sure, could you please link them. I just wanted to avoid outside changes as much as possible for this branch. And currently could not find the working method.

xgui3783:
left TODOs:

- [ ] migrate tools to modular classes
  - [ ] point
  - [ ] line
- [ ] render annotations in side bar
- [ ] remove annotation.service and most code in annotationMode.component

xgui3783:
re: double click/right click cancel editing, is there a precedent for this behaviour? like photoshop or inkscape that uses this shortcut?

fsdavid:
> re: double click/right click cancel editing, is there a precedent for this behaviour? like photoshop or inkscape that uses this shortcut?

Sure for double click. You can check `boxy-svg.com`, `editor.method.ac` and I believe many many more. For me, the normal behaviour for finishing the same type of `drawing` is by double click (It is not necessary to be a real double click I believe we can finish drawing by clicking on the last or the first point).  Right-click is Timos suggestion.

xgui3783:
> Sure for double click. You can check boxy-svg.com, editor.method.ac

The DblCl behaviour are incredible inconsistent between these two applications.

- boxy-svg adds last point, and but does not complete the polyline
- editor.method.ac completes the last point, and completes the polyline
- for editor.methods.ac, click on any previously added point also completes polyline
- I cannot find an example where RightCl completes the polyline (we should serious consider if right click should be a mode of interaciton)

We should probably ask some feedback from sebastian and andrea.

fsdavid:
@xgui3783 Sorry but I could not understand what do you mean. I just googled first SVG editor apps and what I have able to test all of them I could finish `blazier` or any `pan` paintings with point

> > Sure for double click. You can check boxy-svg.com, editor.method.ac
> 
> The DblCl behaviour are incredible inconsistent between these two applications.
> 
> * boxy-svg adds last point, and but does not complete the polyline
> * editor.method.ac completes the last point, and completes the polyline
> * for editor.methods.ac, click on any previously added point also completes polyline
> * I cannot find an example where RightCl completes the polyline (we should serious consider if right click should be a mode of interaciton)
> 
> We should probably ask some feedback from sebastian and andrea.

Sorry but I do not understand what do you mean by `complete the polyline` and sorry if examples were not enough, I just take the first examples in google and send them to you.

Now, I've tried more apps and websites and find out that, mostly double click is a default way to end painting polyline or bazier (click on the first point is also frequent way), and `esc` removes the full editing polygon. 

App: `inkscape`, `Paint S`, `Paint pad light`. 
Web: `https://web.autocad.com/`, `https://www.photopea.com/` - Try `Polygonal Lasso Select`,  `https://pixlr.com/e/#editor`, `https://gojs.net/latest/extensions/PolygonDrawing.html`, `https://www.gifgit.com/image/pen-tool` , `https://roomstyler.com/`

I also tried Adobe XD where double click is not a default way to finish painting, but I guess because double click has another feature which is logical.

I'm still installing some apps to test but for now, my PC went crazy and stucked :d

fsdavid:
No, I'm wrong with `https://web.autocad.com/`. because it is really very slow on my PC. I tried `ARC` and double click was working but I found `pline` where it finishes with `Esc` and `Context menu`. I'm not sure why because it is very slow and I could not test well.

xgui3783:
I mean, the fact that we are discussing what is the best way to end a poly line drawing tells me that anything other than escape is unintuitive.

PR #960, opened 2021-06-28
Quick tour changes

fsdavid:


src/ui/quickTour/quickTourThis.directive.ts
```diff
@@ -14,7 +15,8 @@ export class QuickTourThis implements OnInit, OnChanges, OnDestroy {
   @Input('quick-tour-position') position: TQuickTourPosition
   @Input('quick-tour-overwrite-position') overwritePosition: IQuickTourOverwritePosition
   @Input('quick-tour-overwrite-arrow') overWriteArrow: TemplateRef<any> | string
-  
+  @Input('quick-tour-check-auto-start') quickTourCheckAutoStart: boolean
```
xgui3783, line 18: Probably shouldn't be in this directive

src/ui/quickTour/quickTourThis.directive.ts
```diff
@@ -1,6 +1,7 @@
 import { Directive, ElementRef, Input, OnChanges, OnDestroy, OnInit, TemplateRef } from "@angular/core";
 import { QuickTourService } from "src/ui/quickTour/quickTour.service";
 import { IQuickTourOverwritePosition, TQuickTourPosition } from "src/ui/quickTour/constrants";
+import {LOCAL_STORAGE_CONST} from "src/util/constants";
```
xgui3783, line 4: Probably should not be in this directive. Maybe in service?

src/ui/quickTour/quickTourThis.directive.ts
```diff
@@ -29,6 +31,14 @@ export class QuickTourThis implements OnInit, OnChanges, OnDestroy {
 
   ngOnInit() {
     this.quickTourService.register(this)
+
+    if (this.quickTourCheckAutoStart) {
```
xgui3783, line 35: Ditto above, probably should be in service?

src/ui/quickTour/quickTourThis.directive.ts
```diff
@@ -29,6 +31,14 @@ export class QuickTourThis implements OnInit, OnChanges, OnDestroy {
...
+      if (!localStorage.getItem(LOCAL_STORAGE_CONST.QUICK_TOUR_VIEWED)) {
+        this.quickTourService.startTour()
```
xgui3783, line 37: Also, shouldn't start tour without explanation. Give modal, ask if user wants a quick tour

src/ui/quickTour/quickTourThis.directive.ts
xgui3783, line 4: Also, did you use the wrong constant?

xgui3783 (changes requested):

src/ui/quickTour/quickTourThis.directive.ts
fsdavid: Which approach would you suggest we could use. We need to start a quick tour automatically, but only after the first ordered directive will be registered. So I decide to keep a quick tour service module local and add the input parameter `quick-tour-check-auto-start`.

src/ui/quickTour/quickTourThis.directive.ts
fsdavid: Sorry could not udnerstand

src/ui/quickTour/quickTourThis.directive.ts
fsdavid: I thought the logic should be when the user first visits the app, the tour will automatically appear, would not the modal be unnecessary? turning off quick tour anyways needs a single click.

src/ui/quickTour/quickTourThis.directive.ts
xgui3783: Having a request modal also solves this issue, no? We will have time to have all quick tour directive initialise

src/ui/quickTour/quickTourThis.directive.ts
xgui3783: Hmm. It's very jarring (to me, anyway) if a quick tour is started without explanation.

But let's gather what existing apps do.

fsdavid: From what I remember, in most applications, it is not started automatically at all, but it starts without modal in some apps. To be honest I do not remember any apps with modal asking start quick tour but I believe they exist.

src/ui/quickTour/quickTourThis.directive.ts
fsdavid: Yes, it could.

src/ui/quickTour/quickTourThis.directive.ts
xgui3783: Would be ideal to have something to compare to. 

Can we list some applications ( preferably web ones) and compare?

fsdavid: Could not remember exactly what applications have a quick tour. Do you remember any?

fsdavid: maybe check https://zurb.com/playground/jquery-joyride-feature-tour-plugin

src/ui/quickTour/quickTourThis.directive.ts
xgui3783: aww, my bad. I was in a rush didn't see how LOCAL_STORAGE_CONST was used.

src/ui/quickTour/quickTourThis.directive.ts
xgui3783: I think that's not a good example. The product it sells is targeted at developers, and not users. 

I am looking at https://www.appcues.com/blog/the-5-best-walkthrough-examples and it seems most (at least 3/5) shows an onboarding message (e.g. Would you like to take a product tour > Go > No)

have you asked Hiba perhaps? she might know what might be a good pattern/anti-pattern?

fsdavid: I've also tested apps from this URL and what I've found. As I see:
asana **does not** do anything after clicking get started we should click help and we can get help from the sidebar.
image

Humanity - is showing tooltips while using the app automatically There is not a quick tour, it is more tooltips but **automatic**. 

Pinterest - shows tooltips **automatically**. It marks a special place and adds toast for description
image

Autopilot - **Does not** show anything. They have a creative way of teaching 
image

GoToWebinar - the only one that I found shows **modal** before starting the tour, I and guess the reason is that users should navigate between different pages while touring.

Finally, I agree with you because Humanity and  Pinterest do not have tours like us. They are displaying toasts or tooltips automatically while using the application. I will make some changes to show the modal.

xgui3783: Hmm... it really seems a stepwise quick tour is going out of fashion. 

I really wonder why that is.

fsdavid:
@xgui3783 Welcome message has been added but could not remove `quick-tour-check-auto-start` input parameter of the directive, as this parameter is used to determine whenever viewer is loaded or not. Do you have any idea how could be better to find when the viewer will be loaded and then show the modal?

xgui3783:
in principle, quick tour **can** be launched without viewer loading. 

We really shouldn't try to auto launch quick tour based on whether the viewer is loaded.

fsdavid:
> in principle, quick tour **can** be launched without viewer loading.
> 
> We really shouldn't try to auto launch quick tour based on whether the viewer is loaded.

Just quick tours usually appearing after the user logs in to the application and already sees the app content. in our case user needs to select atlas first to see the app. Especially, our tour is not action base (as it was in the first dev version), it is presentational and just shows the places, so I think it is necessary first to load or select an atlas and only then show tour. I was thinking about `selectedTemplate` subscription into the quick tour service, and we can `take(1)`, but I do not know if it is the better way. what do you think?

xgui3783:
In principle I know what you mean.

But in practice, putting the attach auto-start input to quicktour this makes no sense. 

The best compromise maybe start the auto start flow when at least one quick tour this component has been attached.

fsdavid:
> In principle I know what you mean.
> 
> But in practice, putting the attach auto-start input to quicktour this makes no sense.
> 
> The best compromise maybe start the auto start flow when at least one quick tour this component has been attached.

yeah before I push the changes, I try the same but seems the first directive is registering too early (on the splash screen) and it makes no sense. The result was: 
There are 2 ways the user opens a app: 1 - user goes to the website landing page and 2 - user directly goes to the selected template page. at the 1 way, the dialog is shown and it makes no sense. And 2 - here user sees the first splash screen with white dialog, and after several seconds same dialog changes to a black color and it did not look nice.

xgui3783:
> yeah before I push the changes, I try the same but seems the first directive is registering too early 

Either these quick tours needs to be conditionally rendered when viewer renders, or be attached to other elements, IMO then.

fsdavid:
> > yeah before I push the changes, I try the same but seems the first directive is registering too early
> 
> Either these quick tours needs to be conditionally rendered when viewer renders, or be attached to other elements, IMO then.

What do you think to simply watch on the selected template before we show the modal?

xgui3783:
which components have quick tour this attached, before viewer is instantiated?

fsdavid:
The last one, plugins and tools icons one and one for selected parcellation chips

xgui3783:
the last one can be fixed, so that it will only show when a viewer is present, but that is less relevant.

select parcellation chips - this should **not** be init when viewer is not init'ed. the chip holder is not even visible.

plugins and tools icon:

- should plugin/tool UI be visible when viewer has not yet been initiated?
- if so, should there be a quick tour step for it?

fsdavid:
> * should plugin/tool UI be visible when viewer has not yet been initiated?
> * if so, should there be a quick tour step for it?

I think the answer to both questions are yes

xgui3783:
Then, to me, it makes logical sense that the auto start should start, with a single quick tour card pointing at the tool/plugin

Hmm, but perhaps we could introduce an input `quick-tour-severity`: 'low' | 'medium' | 'high' (probably use proper Enum)

if the only registered quick tour direcitves are of low severity, auto start will not trigger. etc etc

xgui3783:
in this case, both the tool/icon quick tour & end card can be 'low' in severity. the rest being default severity ('medium')

fsdavid:
Nice solution. New commit should contain severity.

xgui3783:
I have some minor tweaks to make. But I also think this is a good solution.

PR #961, opened 2021-07-14
Fix visual bugs

fsdavid:


fsdavid:
I've divided PR into 3 separate PR's

PR #966, opened 2021-07-15
Modify explore in other templates for siibra api

fsdavid:
@xgui3783 Before we merge, the current approach is taken from the existing one. We are using `availableIn` field but still we are searching regions from the `fetchedTemplates`. 
As I mentioned, currently most regions do not have a centroid position, and look like it makes no sense to have `mni152 left` and `mni 152 right` into the `bigbrain` while exploring. 
The current approach works and of course we can use it (it almost never navigates after selecting exploring template) but I suggest 2 (maybe better) approaches.
 1 - keep a current list (e.g. `mni152 left` and `mni 152 right` when `bigbrain` is selected) and when the user clicks for example `mni152 left`, generate a new URL with selected template and selected region and open as a black page.
2 - simplify the list in Big Brain (e.g. have only `mni152` when `bigbrain` is selected) and do not navigate to the region after selecting a template (transformation coordination backend will work). this approach will make code more reliable.

What do you think?

### DIFF
--- a/common/util.js
+++ b/common/util.js
@@ -21,15 +21,15 @@
   }
 
   const HEMISPHERE = {
-    LEFT_HEMISPHERE: `left hemisphere`,
-    RIGHT_HEMISPHERE: `right hemisphere`
+    LEFT_HEMISPHERE: `left`,
+    RIGHT_HEMISPHERE: `right`
   }
 
   exports.getRegionHemisphere = region => {
     if (!region) return null
-    return (region.name && region.name.includes('- right hemisphere') || (!!region.status && region.status.includes('right hemisphere')))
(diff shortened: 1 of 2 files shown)


xgui3783:
can you

1/ fix the unit test and
2/ make the branch name shorter

deploy is failing because the branch name is too long

xgui3783:
re: your alternatives

1/ is a definitely no go. it should be possible for us to do everything in a single page
2/ I thought this is how explore in other template always worked?

fsdavid:
> 2/ I thought this is how explore in other template always worked?

Actually, we had `mni152 left` and `mni152 right` because after changing the template from there, user was navigated to the region.

xgui3783:
can't we do that here?

we can get centroids for mni152 and colin left/right, just not big brain

PR #967, opened 2021-07-15
Dev eiot api

fsdavid:


fsdavid:
> can't we do that here?
> we can get centroids for mni152 and colin left/right, just not big brain

Then it should work now.

PR #968, opened 2021-07-15
Modify explore in other templates for siibra api

fsdavid:
> can't we do that here?
> we can get centroids for mni152 and colin left/right, just not a big brain

@xgui3783 Then it should work with the current way. Could be great to get them.

### DIFF
--- a/src/atlasComponents/parcellationRegion/region.base.ts
+++ b/src/atlasComponents/parcellationRegion/region.base.ts
@@ -4,7 +4,7 @@ import { uiStateOpenSidePanel, uiStateExpandSidePanel, uiActionShowSidePanelConn
 import { distinctUntilChanged, switchMap, filter, map, withLatestFrom } from "rxjs/operators";
 import { Observable, BehaviorSubject, combineLatest } from "rxjs";
 import { ARIA_LABELS } from 'common/constants'
-import { flattenRegions, getIdFromFullId, rgbToHsl } from 'common/util'
+import { flattenRegions, getIdFromFullId, getIdFromKgIdObj, rgbToHsl } from 'common/util'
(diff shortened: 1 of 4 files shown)


src/atlasComponents/parcellationRegion/region.base.ts
```diff
@@ -264,59 +264,77 @@ export class RenderViewOriginDatasetLabelPipe implements PipeTransform{
...
-    
+    // atlas is the single source of truth
+
     const otherTemplates = fetchedTemplates
-      .filter(({ fullId }) => getIdFromFullId(fullId) !== selectedTemplateId)
-      .filter(({ ['@id']: id, fullId }) => atlasTemplateSpacesIds.includes(id || fullId))
+      .filter(({ ['@id']: id }) => id !== regionOfInterest.context.template['@id'])
+      .filter(({ ['@id']: id }) => atlasTemplateSpacesIds.includes(id))
+      .filter(({ ['@id']: id }) => regionOfInterest.availableIn.map(ai => ai.id).includes(id))
+
```
xgui3783, line 281: can you pull these 3 into a single filter call?

xgui3783 (changes requested):
Just a minor fix.

xgui3783:
also, another note. 

you updated `e2e/protractor.conf.js`, even though no material change has been made. I know your IDE did formatting, but it clutters the commit, as well as the history. For a reviewer, it's bad form for them to scroll through tens if not hundred lines of formatting changes.

It is always a good idea to check what is being committed/added, instead of blindly doing `git add .` or `git commit -a`.

I recommend before `git add` or `git commit -a`, go through the updated file(s), and check for:

- if any secrets have been added by mistake (including tokens)
- remove any unused imports/console.log/debugger (I think linter can eventually help here)
- unstage/reset any changes to file that are unsubstantial

fsdavid:
You are right, I understand. Usually, I always try not to have that formatting thing, but yesterday I missed it.

PR #969, opened 2021-07-16
Fix show connectivity map over non mixable layers

fsdavid:


xgui3783:
I am aware of this issue, but I don't think this is the correct approach.

I will investigate, and when fixed, will leave a comment here.

fsdavid:
What do you think is the correct approach. I could help.
You could think about the current approach. If there is not a case when non-mixable layer has priority over clearViewFlag (I believe so), maybe the current approach is correct.

xgui3783:
superceded by f9d10c9

PR #1009, opened 2021-09-06
Dev reg hie coll

fsdavid:


xgui3783:
Can you be more specific which issue(s) this draft PR is addressing?

fsdavid:
> Can you be more specific which issue(s) this draft PR is addressing?

Sure

xgui3783:
Can you update the draft PR name to something that makes sense, too?

Also, wouldn't it make sense to update the flat tree rather than create a new component?

fsdavid:
I will change the name of PR before the merge. I tried to make a name as short as possible because of the 
Angular material already has a module for the tree and it makes it simpler, easy to understand, and has many useful features to use. As I see flat tree component is made before angular material integration. I do not know if we need a tree component anymore, if not I can clean the code. Full code is not simplified and cleared yet, because first I thought to look at the prototype and get feedback.

fsdavid:
@xgui3783 Do you have any idea why the docker build does not work?

xgui3783:
It's the build process failing? I thought it's the pushing of image to docker registry that was the issue? 

The registry issue should be fixed. Just rerun the pipeline and it should work.

fsdavid:
Yes, I mean docker registry. Thank you very much.

PR #1021, opened 2021-09-17
Fix space key behaviour on full view

fsdavid:


xgui3783:
Hmm, I will have to think about this briefly.

I am not too happy on how the spacebar listener was implemented.

fsdavid:
What do you think to merge PR as the fix of the production and change the logic in the dev version? The issue happens not only on annotation mode but any input text, like both region searches.

xgui3783:
I briefly remember implementing fix so that region search does not result in cycling view.

What I don't want is to have 2 solutions for the same problem.

fsdavid:
As I see production version still has bugs for region searches both on the sidebar and region hierarchy. Also as you mentioned it is the global problem of how spacebar listener works. I think implementing fixes of each separate component would not be the best way in such kind of situation.

xgui3783:
Hmm, that is true. There had been regressions.

But I think rather than checking `(ev.target as HTMLElement).classList.contains('neuroglancer-panel')`, we should check if ev.target === input // text area

I feel very hesitant to hard code class. 

at the same time, I wonder if the event listener should be in the service at all.

xgui3783:
don't make any edits yet, I need to review how it is implemented currently

fsdavid:
Ok. If you will be agree I could change code with ```!['input', 'textarea'].includes((ev.target as HTMLElement).tagName.toLowerCase())```

fsdavid:
@xgui3783 What do you think may I use some different approaches to make a quick fix for tomorrow's workshop?

xgui3783:
In lieu with the urgency of the issue, let's merge this for now.

we can undo and revisit this issue at a later date.

PR #1075, opened 2021-10-19
Reg hiera merged

fsdavid:


fsdavid:
Updated with: https://github.com/FZJ-INM1-BDA/siibra-explorer/pull/1091

Issue #1008, opened 2021-09-02
from support email feedback: allow selection of multiple regions


xgui3783:
(support ticket #8753)

possibly related to two separated, but related feature requests:
- [ ] select super structure
- [ ] select arbitrary structures on (potentially) different branches

xgui3783:
More detail:

Able to visualise spatial relationships between two regions and their spatial relationship.

visualise the connection between the regions (require support from siibra-python and siibra-api)

dickscheid:
I second that his is an important next feature. siibra-python can handle multi-region selections in principle already, but it is not yet obviously modelled on the surface.

xgui3783:
also requested by @ch-schiffer

dickscheid:
has been merged into new issue

PR #1091, opened 2021-11-02
Update region hierarchy with angular material

fsdavid:


xgui3783:
the tree is slightly misaligned (by maybe 0.5 character)

current:
Screenshot from 2021-11-02 14-54-20

this update:

Screenshot from 2021-11-02 14-53-21

src/atlasComponents/parcellation/regionHierachy/region-tree/region-tree.component.html
```diff
@@ -0,0 +1,40 @@
...
+    <ng-container *cdkVirtualFor="let node of dataSource; let i = index;">
+        <!-- Note that the [style.padding-left] is essentially what cdkTreeNodePadding is doing under the hood -->
+        <div class="d-flex" [style.height]="lineHeight + 'px'">
+
+            <div *ngFor="let level of [].constructor(node.level); let last = last; let j = index"
+                 [style.height]="lineHeight + 'px'"
+                 [ngStyle]="{borderBottom: last? '1px dashed' : '',
+                             borderLeft: node['missBorderPaintingLevels']
+                                && node['missBorderPaintingLevels'].includes(j+1)? '' : '1px dashed',
+                             marginTop: - lineHeight/2 + 'px'}"
```
xgui3783, line 12: please use a pipe for this

src/atlasComponents/parcellation/regionHierachy/region-tree/region-tree.component.html
```diff
@@ -0,0 +1,40 @@
...
+                 style="width: 20px">
+            </div>
+
+            <div *ngIf="node.hasChildren; else lastChildTmp">
+                <button mat-button (click)="toggleNode(node)"
+                        [style.height]="lineHeight + 'px'"
```
xgui3783, line 18: ditto above

src/atlasComponents/parcellation/regionHierachy/region-tree/region-tree.component.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+  beforeEach(() => {
+    fixture = TestBed.createComponent(RegionTreeComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+});
```
xgui3783, line 25: ideally there should be more tests than just `should create`

src/atlasComponents/parcellation/regionHierachy/region-tree/region-tree.component.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  @Input() regions: any
+
+  @Input() public inputItem: any = [{
+    name : 'Untitled',
+    children : [],
+  }]
+  @Input() public childrenExpanded: boolean = true
+
+  public _searchString: string
```
xgui3783, line 23: set to private? no need to be public i think?

src/atlasComponents/parcellation/regionHierachy/region-tree/region-tree.component.ts
```diff
@@ -0,0 +1,139 @@
...
+  
+  toggleNode(node) {
+    if (this.treeControl.isExpanded(node)) {
+      this.treeControl.collapseDescendants(node)
+    } else {
+      this.treeControl.toggle(node)
+    }
+  }
+
+  filterChanged(filterText: string) {
```
xgui3783, line 47: this function in general is very confusing. 

I know what it tries to do at the very end, but perhaps we can clean it up a little bit

src/atlasComponents/parcellation/regionHierachy/region-tree/region-tree.component.ts
```diff
@@ -0,0 +1,139 @@
...
+    event.preventDefault()
+    if (JSON.stringify(node) === JSON.stringify(this.selectedRegion)) {
+      this.selectedRegion = {}
+    } else {
+      this.selectedRegion = node
+    }
+    this.selectRegion.emit(node)
+  }
+
+}
```
xgui3783, line 139: I think I know why we keep a reference to `selectedRegion`, as it will two way bind to the selected region. 

Still, I think it's less brittle if we keep a single source of truth, and inject selected region as an input, or get it from the store (if the store is injected already).

It would also simplify `toggleSelection` method

src/atlasComponents/parcellation/regionHierachy/regionHierarchy.template.html
```diff
@@ -26,16 +26,20 @@
...
         class="tree-body">
-        <flat-tree-component
-          (treeNodeClick)="handleClickRegion($event)"
-          (totalRenderedListChanged)="handleTotalRenderedListChanged($event)"
-          [inputItem]="aggregatedRegionTree"
-          [renderNode]="displayTreeNode"
-          [searchFilter]="filterTreeBySearch">
-          
-        </flat-tree-component>
+
```
xgui3783, line 29: can you remove all references to `flat-tree-component`?

I don't think it's used anywhere else but here.

src/components/components.module.ts
```diff
@@ -78,6 +78,7 @@ import { TileCmp } from './tile/tile.component';
 
     TreeSearchPipe,
     TreeBaseDirective,
+    HighlightPipe,
```
xgui3783, line 81: I didn't think highlight pipe is needed in `component.module`?

src/atlasComponents/parcellation/regionHierachy/region-tree/region-tree.component.html
```diff
@@ -0,0 +1,40 @@
...
+                <div class="d-flex align-items-center pl-4">
+                    <mat-radio-button [value]="node" (click)="toggleSelection($event, node)" class="align-self-end">
+                        <span [innerHTML]="node.name | highlightPipe: searchString"></span>
+                    </mat-radio-button>
+<!--                    <button mat-icon-button-->
+<!--                            (click)="navigateToRegion.emit(node)"-->
+<!--                            [style.height]="lineHeight + 'px'">-->
+<!--                        <mat-icon fontSet="fas" fontIcon="fa-map-marked-alt">-->
+<!--                        </mat-icon>-->
+<!--                    </button>-->
```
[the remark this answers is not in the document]
fsdavid: This is the navigate button. As I know, we will get region positions easier, we will just uncomment. I will add ToDo

src/atlasComponents/parcellation/regionHierachy/region-tree/region-tree.component.html
fsdavid: Using a directive would be acceptable?

fsdavid:
@xgui3783 could you please have a look? I've made the changes.

xgui3783:
will look more closely at the code later, but immediately, 2 things jump out at me:

Screenshot from 2021-12-03 17-05-36

- the tree is offset (see https://github.com/FZJ-INM1-BDA/siibra-explorer/pull/1091#issuecomment-957627022)
- I really do not like the selected region, how it is bordered. Why can't we still have selected = yellow text?

fsdavid:
> * the tree is offset 
Sorry, what do you mean with offset? space between the collapse arrow and the buttons?
> * I really do not like the selected region, how it is bordered. Why can't we still have selected = yellow text?
Sure we can have yellow color for selected regions

xgui3783:
> Sorry, what do you mean with offset? space between the collapse arrow and the buttons?

offset between arrow and the tree trunk and arrow. See screenshot below.

highlight_tree

xgui3783:
still have to review the code, sorry, didn't get a chance to do that.

but re: tree branches, it's now poking out from the bottom =( 

also, I would reconsider single click on tree branch == collapse/expand tree. We want to reintroduce selection of higher hierarchy regions in the future (see #1008 ) 

re: go there icon in tree view:
- it sometimes does not work
- it doesn't look good at all

I say let's take it out for the time being

fsdavid:
> also, I would reconsider single click on tree branch == collapse/expand tree. We want to reintroduce selection of higher hierarchy regions in the future (see #1008 )

Yes, I agree. After multi-selection, we can use checkboxes. Then single click will select parent regions and collapse/expand will happen only on arrow click (when checkbox will be there, it will very clear that single click will select the region instead of collapsing).

xgui3783:
> Yes, I agree. After multi-selection, we can use checkboxes. Then single click will select parent regions and collapse/expand will happen only on arrow click (when checkbox will be there, it will very clear that single click will select the region instead of collapsing).

As such, I don't think we should introduce a different interaction to the parent region.

src/atlasComponents/parcellation/module.ts
```diff
@@ -25,7 +29,11 @@ import { GetParcPreviewUrlPipe } from "./getParcPreviewUrl.pipe";
     RegionTextSearchAutocomplete,
 
     FilterNameBySearch,
+    RegionTreeComponent,
```
xgui3783, line 32: I am half half on naming this component something generic and keep it as something like `SearchableTreeModule` or s th

src/atlasComponents/parcellation/regionHierachy/region-tree/region-tree.component.html
```diff
@@ -0,0 +1,43 @@
...
+                        <span [innerHTML]="node.name | highlightPipe: searchString"
+                              [ngClass]="selectedRegion && selectedRegion.name === node.name? 'regionSelected' : ''"
+                              (click)="toggleSelection($event, node)"></span>
+                        </button>
+
+                        <small>
+                            <mat-icon fontSet="fas" fontIcon="fa-map-marked-alt"
+                                      class="cursor-pointer ml-2 tree-item-height"
+                                      (click)="navigateToRegion.emit(node)">
+                            </mat-icon>
```
xgui3783, line 36: see comment about `go to` region

the icon is not very pleasant, and half o the region doesn't work. 

I say we remove this

src/atlasComponents/parcellation/regionHierachy/region-tree/region-tree.component.html
```diff
@@ -0,0 +1,43 @@
...
+            <div *ngIf="node.hasChildren; else lastChildTmp" class="arrow-alignment" >
+                <button mat-button (click)="toggleNodeExpansion(node)"
+                        [attr.aria-label]="'Toggle ' + node.name"
+                        class="d-flex align-items-center tree-item-height pl-0">
+                    <i [ngClass]="!treeControl.isExpanded(node) ? 'r-270' : ''" class="fas fa-chevron-down mr-1"></i>
+                    <span [innerHTML]="node.name | highlightPipe: searchString"></span>
+                </button>
+            </div>
+            <ng-template #lastChildTmp>
+                <div>
```
xgui3783, line 19: why do we need a naked div here?

src/atlasComponents/parcellation/regionHierachy/region-tree/region-tree.component.html
```diff
@@ -0,0 +1,43 @@
...
+            >
+            </div>
+
+            <div *ngIf="node.hasChildren; else lastChildTmp" class="arrow-alignment" >
+                <button mat-button (click)="toggleNodeExpansion(node)"
+                        [attr.aria-label]="'Toggle ' + node.name"
+                        class="d-flex align-items-center tree-item-height pl-0">
+                    <i [ngClass]="!treeControl.isExpanded(node) ? 'r-270' : ''" class="fas fa-chevron-down mr-1"></i>
+                    <span [innerHTML]="node.name | highlightPipe: searchString"></span>
+                </button>
```
xgui3783, line 16: I am not too sure of using button. I feel it's closer to a list rather than a button (with hoverable/clickable state, that is)

src/atlasComponents/parcellation/regionHierachy/region-tree/region-tree.component.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  public treeControl: FlatTreeControl<any> = new FlatTreeControl<any>((e => e.level), (e => e.hasChildren))
+  public dataSource: MatTreeFlatDataSource<any, any>
+  public lineHeight = 20
+
+
+
+  public treeFlattener =
+    new MatTreeFlattener<any, any>(
+      ((node: any, level: number) => ({
```
xgui3783, line 41: can you type the node object here?

src/atlasComponents/parcellation/regionHierachy/region-tree/region-tree.component.ts
```diff
@@ -0,0 +1,142 @@
...
+        const children = object.children.reduce(getChildren, [])
+        if (children.length) result.push({ ...object, children })
+      }
+      return result
+    }
+
+    return array.reduce(getChildren, [])
+  }
+
+  correctToggledItems() {
```
xgui3783, line 96: I see you extracted the function here, but I still struggle to understand what this function is suppose to do. 

is it suppose to add `missBorderPaintingLevels` property and nothing else?

also, (relevant generally), these functions should be private, as it is not called outside the component.

src/atlasComponents/parcellation/regionHierachy/region-tree/region-tree.component.ts
```diff
@@ -0,0 +1,142 @@
...
+    this.dataSource.data = filteredItems
+
+    this.correctToggledItems()
+
+    this.treeControl.expandAll()
+  }
+
+
+
+  filter = (array, text) => {
```
xgui3783, line 73: should be made private

and please add typing

src/atlasComponents/parcellation/regionHierachy/region-tree/region-tree.component.ts
```diff
@@ -0,0 +1,142 @@
...
+              node.level
+            ] : [node.level]
+          }
+        }
+      }
+    }
+  }
+
+  toggleSelection(event, node) {
+    event.preventDefault()
```
xgui3783, line 138: why is preventdefault needed here? (I genuinely do not know/forgot)

src/environments/environment.common.ts
```diff
@@ -5,7 +5,7 @@ export const environment = {
   PRODUCTION: true,
   BACKEND_URL: null,
   DATASET_PREVIEW_URL: 'https://hbp-kg-dataset-previewer.apps.hbp.eu/v2',
-  BS_REST_URL: 'https://siibra-api-latest.apps-dev.hbp.eu/v1_0',
+  BS_REST_URL: 'https://siibra-api-stable.apps.hbp.eu/v1_0',
```
xgui3783, line 8: when it gets deployed on dev cluster, (I think) it gets overwritten to be latest.

I don't mind on our dev machine it targets stable, but just keep in mind, we may get a lot of "it works on my machine" issue down the line.

src/sharedModules/angularMaterial.module.ts
```diff
@@ -60,6 +62,8 @@ const defaultDialogOption: MatDialogConfig = new MatDialogConfig()
     ScrollingModule,
     MatToolbarModule,
     ClipboardModule,
+    MatTreeModule,
+    MatRadioModule
```
xgui3783, line 66: we are no longer using radio? please remove

src/util/util.module.ts
```diff
@@ -48,10 +47,10 @@ import { MergeObjPipe } from "./mergeObj.pipe";
     DoiParserPipe,
     GetFilenamePipe,
     CombineFnPipe,
+    SafeStylePipe,
```
xgui3783, line 50: we can probably remove safestyle pipe, it is not used anywhere

xgui3783 (changes requested):
mostly small cosmetic changes, except the big block of code in tree component

I can't help but to think there must be an easier way to deal with the children

src/atlasComponents/parcellation/regionHierachy/region-tree/region-tree.component.ts
fsdavid: It is a region object. Do we have some interface or type for the region?

xgui3783:
I figured as much.

Don't worry about typing for the time being, but do try to:

- rename component so it's agnostic (remove all reference to region)
- move to its own module

I can/will take a look at the implementation / write interfaces once you
are done

src/atlasComponents/parcellation/regionHierachy/region-tree/region-tree.component.ts
fsdavid: Yes, I'll change name to `addmissBorderPaintingLevels`

xgui3783:
superceded by #1169 specifically e0088c090cf5890d0f0def47170326c7d573fb32

Issue #1042, opened 2021-10-06
'q' will hide all layers except for template


dickscheid:
The 'q' key is used to hide segmentations / maps. This is especially important when adding annotatinos in zoomed-in bigbrain views. However, pressing 'q' will also hide the user annotations. It would be great if user annotations are behind a different layer, and not subject to this shortcut key. They should always be shown when annotation mode is active.

xgui3783:
Currently, the viewer will toggle all layers except for the first one. This also affects local nifti volume, for example.

A more definitive approach is to keep track of all segmentation layer(s), and toggle the visibility.

fsdavid:
I see now, yes, I thought that list will need to update by hand. Sure I will make the change.

xgui3783:
fixed by 2.6.0

PR #1092, opened 2021-11-02
Do not hide annotations with parcellation

fsdavid:


xgui3783:
does this PR deal with the `q` short cut?

fsdavid:
Yes, I will mention it

fsdavid:
Keeping a list of all segmentation ng volumes would not be too buggy for the feature? 

I thought, while hiding parcellation, we have 3 ways to make it done, 1 - keep a list of the layers that should be hidden, 2 - layers that should not, 3 - simply filter annotation layer. As into the 1 approach, the list could be changed and is not static, and on another hand 2 and 3 approaches will use static data (annotation layer will always have the same name and will not depend on the spaces, templates, or parcellations.) I decided to choose from 2 and 3. Between these two, I choose the third one because I do know and didn't think that we will have some other layer types which will need to stay displayed when parcellation hides. I want also to mention that change is made into the method which is used to hide the parcellation which is already concrete enough and adding a filter for annotation would not be confusing.

xgui3783:
I advocate for 1/ . see https://github.com/FZJ-INM1-BDA/siibra-explorer/issues/1042#issuecomment-957587934

We don't know which other layers will be added in the future (good example is for example neuromorphology mesh, which will also be hidden)

> As into the 1 approach, the list could be changed and is not static, 

that is fine. That's what rxjs is for.

fsdavid:
I see now, yes, I thought that list will need to update by hand. Sure I will make the change.

src/viewerModule/nehuba/viewerCtrl/viewerCtrlCmp/viewerCtrlCmp.component.ts
```diff
@@ -156,6 +161,7 @@ export class ViewerCtrlCmp{
     const visibleParcLayers = ((window as any).viewer.layerManager.managedLayers)
       .slice(1)
       .filter(({ visible }) => visible)
+      .filter(l => Object.keys(this.pureConstantService.ngLayerObj[this.selectedTemplateId]).includes(l.name))
```
xgui3783, line 164: can you have a guard in case nglayerobj[key] is undefined?

xgui3783 (changes requested):
can you please also fix the test?

---

I am still not 100% sold on the monkey patch of `this.NgLayerObj` in pureconstant service.

I would, say, make NgLayerObj private, but implement a public method `getNgSpec(atlasId: string, templateId: string)`.

The method should return:

```ts
{
  template: NgSpec,
  segmentations: NgSpec
}
```

where `NgSpec` is defined as:

```ts
type NgLayerObj = {
  [key: string]: {
    source: string,
    type: 'segmentation' | 'image',
    ... (etc)
  }
}
```

fsdavid:
I've pushed commit with little changes from your comment.
1 - Instead of `getNgSpec(atlasId: string, templateId: string)` I've used `getNgSpec(templateId: string)` as ngLayers list does not contain atlas
2 - I'm returning only ng layers - not template

src/util/pureConstant.service.ts
```diff
@@ -515,7 +525,12 @@ Raise/track issues at github repo: <a target = "_blank" href = "${this.repoUrl}"
     shareReplay(1)
   )
 
-  public ngLayerObj = {}
+  private ngLayerObj: NgLayerObj = {}
```
xgui3783, line 528: Can you add a TODO here to either improve the readability and/or remove this entirely eventually?

src/util/pureConstant.service.ts
```diff
@@ -515,7 +525,12 @@ Raise/track issues at github repo: <a target = "_blank" href = "${this.repoUrl}"
...
+
+  getNgLayers(templateId: string) {
+    return this.ngLayerObj && this.ngLayerObj[templateId]
+      ? this.ngLayerObj[templateId] : []
+  }
```
xgui3783, line 533: this can probably be simplified to:
```ts
getNgLayers(templateId: string): NgLayerObj[keyof NgLayerObj] {
  return this.ngLayerObj[templateId] || {}
}
```

src/viewerModule/nehuba/viewerCtrl/viewerCtrlCmp/viewerCtrlCmp.component.spec.ts
```diff
@@ -44,6 +44,12 @@ describe('> viewerCtrlCmp.component.ts', () => {
           {
             provide: NEHUBA_INSTANCE_INJTKN,
             useValue: new BehaviorSubject(mockNehubaViewer)
+          },
+          {
+            provide: PureContantService,
+            useValue: {
+              backendUrl: `http://localhost:3000/`
```
xgui3783, line 51: I don't think `backendUrl` needs to be injected at all? if anything it should be `getNgLayers` ?

src/util/pureConstant.service.ts
xgui3783: also, I think this method probably should be called `getNehubaConfigFromTemplateId`

xgui3783 (changes requested):

xgui3783:
> Instead of `getNgSpec(atlasId: string, templateId: string)` I've used `getNgSpec(templateId: string)` as ngLayers list does not contain atlas

this can easily be rectified, IMO. simply add an additional key at the top level of `ngLayerObj`?

src/viewerModule/nehuba/viewerCtrl/viewerCtrlCmp/viewerCtrlCmp.component.spec.ts
fsdavid: Whithout `backendUrl` it returns errors again

xgui3783: that is very strange. I will investigate.

fsdavid:
It should work now

src/viewerModule/nehuba/viewerCtrl/viewerCtrlCmp/viewerCtrlCmp.component.spec.ts
fsdavid: Thanks

src/util/pureConstant.service.ts
```diff
@@ -525,18 +527,21 @@ Raise/track issues at github repo: <a target = "_blank" href = "${this.repoUrl}"
     shareReplay(1)
   )
 
+  //ToDo improve the readability
   private ngLayerObj: NgLayerObj = {}
 
-  getNgLayers(templateId: string) {
-    return this.ngLayerObj && this.ngLayerObj[templateId]
-      ? this.ngLayerObj[templateId] : []
+  getNehubaConfigFromTemplateId(atlasId: string, templateId: string) {
```
xgui3783, line 533: sorry to nitpick, this should be `getNehubaConfigFromAtlasTmplIds` now.

src/viewerModule/nehuba/viewerCtrl/viewerCtrlCmp/viewerCtrlCmp.component.ts
```diff
@@ -156,6 +165,11 @@ export class ViewerCtrlCmp{
     const visibleParcLayers = ((window as any).viewer.layerManager.managedLayers)
       .slice(1)
       .filter(({ visible }) => visible)
+
+      .filter(l => {
+        const layers = this.pureConstantService.getNehubaConfigFromTemplateId(this.selectedAtlasId, this.selectedTemplateId)
```
xgui3783, line 170: also, minor perf improvement ot be gained by moving this out of the filter loop

src/viewerModule/nehuba/viewerCtrl/viewerCtrlCmp/viewerCtrlCmp.component.ts
```diff
@@ -156,6 +165,11 @@ export class ViewerCtrlCmp{
     const visibleParcLayers = ((window as any).viewer.layerManager.managedLayers)
       .slice(1)
```
xgui3783, line 166: this is probably no longer necessary

src/util/pureConstant.service.ts
xgui3783: actually, also, can you make this function async?

src/viewerModule/nehuba/viewerCtrl/viewerCtrlCmp/viewerCtrlCmp.component.ts
fsdavid: Sorry, what do you mean?

xgui3783: The `.slice(1)` removes the first element in an array. 

The original logic goes something like this:

- find all layers that are visible
- toggle the visibility of all **except** the first layer (hence `.slice(1)`)

But now, with `pureService.getNehubaConfigFromAtlasTmplIds`, we no longer have to do this. We can know exactly which layers are parcellations, which layers are image.

fsdavid: Ook thanks.

PR #1126, opened 2021-12-13
Fix perspective orientation button colors

fsdavid:


xgui3783:
Feel free to do the check list, if I don't get a chance. 

I will merge if the checks have been passed by either one of us

fsdavid:
Only the knows issues, which already exist on the production, couldn't pass the test.

PR #1148, opened 2022-02-01
Add viewer tab on settings

fsdavid:


src/ui/config/configCmp/config.component.ts
```diff
@@ -55,11 +63,22 @@ export class ConfigComponent implements OnInit, OnDestroy {
...
 
+  private nehubaViewer: NehubaViewerUnit
+  private get viewer(){
+    return this.injectedViewer || (window as any).viewer
+  }
+
   constructor(
     private store: Store<IavRootStoreInterface>,
     private pureConstantService: PureContantService,
+    @Optional() @Inject(VIEWER_INJECTION_TOKEN) private injectedViewer,
```
xgui3783, line 74: there's no longer need for `VIEWER_INJECTION_TOKEN`

src/ui/config/configCmp/config.component.ts
```diff
@@ -185,4 +204,81 @@ export class ConfigComponent implements OnInit, OnDestroy {
   }
 
   public stepSize: number = 10
+
+  public get axisLineVisible() {
+    const panel: any = Array.from(this.viewer.display.panels)[0]
+    return panel?.viewer?.showAxisLines?.value
+  }
```
xgui3783, line 211: There is an easier method

```
viewer.showAxisLines.value
```

and to set:
```
viewer.showAxisLines.restoreState(true || false)
```

What I am more concerned is this method called on every render cycle, which could make it a big liability.

src/ui/config/configCmp/config.component.ts
```diff
@@ -185,4 +204,81 @@ export class ConfigComponent implements OnInit, OnDestroy {
...
+  public set axisLineVisible(value) {
+    const panel: any = Array.from(this.viewer.display.panels)[0]
+    if (panel && panel.viewer) {
+      panel.viewer.showAxisLines.value = value
+    }
+  }
+
+  public toggleAxisLines(value) {
+    this.axisLineVisible = value
+  }
```
xgui3783, line 222: see above.

src/ui/config/configCmp/config.component.ts
```diff
@@ -185,4 +204,81 @@ export class ConfigComponent implements OnInit, OnDestroy {
...
+  public set showScaleBar(value) {
+    const panel: any = Array.from(this.viewer.display.panels).find((p: any) => p.viewer?.orthographicProjection)
+    if (panel && panel.viewer) {
+      panel.viewer.showScaleBar.value = value
+    }
+  }
+
+  public toggleScaleBar(value) {
+    this.showScaleBar = value
+  }
```
xgui3783, line 239: ditto above with `viewer.showScaleBar`

src/ui/config/configCmp/config.component.ts
```diff
@@ -185,4 +204,81 @@ export class ConfigComponent implements OnInit, OnDestroy {
...
+  public get sliceBackgroundRgb() {
+    const color = this.sliceBackground
+    return '#' + [color[0], color[1], color[2]].map(x => x.toString(16).length === 1 ? '0' + x.toString(16) : x.toString(16))
+      .join('')
+  }
+  
+  public get sliceBackground() {
+    const panel: any = Array.from(this.viewer.display.panels).find((p: any) => p.viewer?.orthographicProjection)
+    return panel?.config?.layout.useNehubaPerspective.drawSubstrates.color
+  }
```
xgui3783, line 251: there is an easier way to get nehuba config

```
(window as any).nehubaViewer.config
```

That being said, we should probably do what we did with nehubaViewer, and inject it as an observable.

e.g. something like

```
@Optional() @Inject(NEHUBA_INSTANCE_INJTKN) nehubaViewer$:
```

src/ui/config/configCmp/config.component.ts
```diff
@@ -185,4 +204,81 @@ export class ConfigComponent implements OnInit, OnDestroy {
...
+    return this.sliceBackground[3] > 0
+  }
+
+  public set showBackground(value) {
+    const panel: any = Array.from(this.viewer.display.panels).find((p: any) => p.viewer?.orthographicProjection)
+    if (panel && panel.config) {
+      panel.config.layout.useNehubaPerspective.drawSubstrates.color[3] = value ? 0.2 : 0
+    }
+    this.nehubaViewer.redraw()
+  }
```
xgui3783, line 275: I think if we set the `drawSubstrates` to falsy, and then call redraw, it would also work (probably more performant)

src/ui/config/configCmp/config.component.ts
```diff
@@ -185,4 +204,81 @@ export class ConfigComponent implements OnInit, OnDestroy {
...
+
+  public toggleShowBackground(value) {
+    this.showBackground = value
+  }
+
+  public hexToRgb = hex => hex.replace(/^#?([a-f\d])([a-f\d])([a-f\d])$/i,(m, r, g, b) => '#' + r + r + g + g + b + b)
+    .substring(1).match(/.{2}/g).map(x => parseInt(x, 16))
```
xgui3783, line 282: do not use SO answer (at least directly). It's CC-SA license, and not compatible with our code. 

There is a hextorgb code in repository (it's in worker, so unfortunately you can't import it, but you can copy paste it)

xgui3783 (changes requested):
In addition, can you merge from `staging` so that the docker build can succeed?

fsdavid:
Moved here:
https://github.com/FZJ-INM1-BDA/siibra-explorer/pull/1150

PR #1153, opened 2022-02-03
Dev view settings

fsdavid:


xgui3783:
Thanks for the new PR @fsdavid .

Do you want me to re add the review from last time?

Also, somewhat related, can you go through https://github.com/FZJ-INM1-BDA/siibra-explorer/branches and delete the branches you don't need?

fsdavid:
Sure, I will clean the branches. I will use review from here https://github.com/FZJ-INM1-BDA/siibra-explorer/pull/1148

src/ui/config/configCmp/config.component.ts
```diff
@@ -185,4 +200,67 @@ export class ConfigComponent implements OnInit, OnDestroy {
...
+  }
+
+  public toggleShowBackground(value) {
+    this.showBackground = value
+  }
+
+  public hexToRgb(hex) {
+    return hex.match(/\w\w/g).map(x => parseInt(x, 16))
+  }
+
```
xgui3783, line 265: this method can be private, it seems

src/ui/config/configCmp/config.component.ts
```diff
@@ -185,4 +200,67 @@ export class ConfigComponent implements OnInit, OnDestroy {
   }
 
   public stepSize: number = 10
+
+  public get axisLineVisible() {
+    return this.viewer.showAxisLines.value
+  }
```
xgui3783, line 206: I have concerns about the performance implications on public getter, especially if it's accessible in the template. 

I think it's safer to setup a local variable to track the axisLineVisible.

the setter method is fine, IMO.

src/ui/config/configCmp/config.component.ts
```diff
@@ -185,4 +200,67 @@ export class ConfigComponent implements OnInit, OnDestroy {
...
+    return this.viewer.showScaleBar.value
+  }
+
+  public set showScaleBar(value) {
+    this.viewer.showScaleBar.restoreState(value)
+  }
+
+  public toggleScaleBar(value) {
+    this.showScaleBar = value
+  }
```
xgui3783, line 227: was scalebar in scope? if not can we remove it?

src/ui/config/configCmp/config.component.ts
```diff
@@ -185,4 +200,67 @@ export class ConfigComponent implements OnInit, OnDestroy {
...
+
+  public get sliceBackgroundRgb() {
+    const color = this.sliceBackground
+    return '#' + [color[0], color[1], color[2]].map(x => x.toString(16).length === 1 ? '0' + x.toString(16) : x.toString(16))
+      .join('')
+  }
+  
+  public get sliceBackground() {
+    return this.nehubaViewer.config.layout.useNehubaPerspective.drawSubstrates.color
+  }
```
xgui3783, line 238: ditto above.

src/ui/config/configCmp/config.component.ts
```diff
@@ -185,4 +200,67 @@ export class ConfigComponent implements OnInit, OnDestroy {
...
+
+  public get showBackground() {
+    return this.sliceBackground[3] > 0
+  }
+
+  public set showBackground(value) {
+    this.nehubaViewer.config.layout.useNehubaPerspective.drawSubstrates.color[3] = value ? 0.2 : 0
+    this.nehubaViewer.redraw()
+  }
```
xgui3783, line 256: You can set `.drawSubstrates` to falsy, I think. 

It should achieve the same result, perhaps more performant (since gl will not even need to try to draw the substrate)

xgui3783 (changes requested):
In addition, can you also please add typing to the methods you introduced?

edit: and also change the PR name to something that makes sense?

fsdavid:
@xgui3783  Should be fixed now

src/ui/config/configCmp/config.component.ts
fsdavid: Setting it simply true/false is changing the default opacity of the background

src/ui/config/configCmp/config.component.ts
```diff
@@ -185,4 +199,67 @@ export class ConfigComponent implements OnInit, OnDestroy {
   }
 
   public stepSize: number = 10
+  private _axisLineVisible = true
+  private _sliceBackground
```
xgui3783, line 203: can you add typing

src/ui/config/configCmp/config.component.ts
```diff
@@ -185,4 +199,67 @@ export class ConfigComponent implements OnInit, OnDestroy {
...
+  }
+
+  private set axisLineVisible(value) {
+    this.viewer.showAxisLines.restoreState(value)
+    this._axisLineVisible = this.viewer.showAxisLines.value
+  }
+
+  public toggleAxisLines(value) {
+    this.axisLineVisible = value
+  }
```
xgui3783, line 218: `toggleAxisLines` should be a function that takes no arguments. Otherwise, it would be called `setAxisLinesVisibility`

src/ui/config/configCmp/config.component.ts
```diff
@@ -185,4 +199,67 @@ export class ConfigComponent implements OnInit, OnDestroy {
...
+
+
+  public get sliceBackgroundRgb() {
```
xgui3783, line 221: can you add typing?

src/ui/config/configCmp/config.component.ts
```diff
@@ -185,4 +199,67 @@ export class ConfigComponent implements OnInit, OnDestroy {
...
+    const color = this.sliceBackground
+    return '#' + [color[0], color[1], color[2]].map(x => x.toString(16).length === 1 ? '0' + x.toString(16) : x.toString(16))
+      .join('')
+  }
+
+  public get sliceBackground() {
```
xgui3783, line 227: can you add typing?

src/ui/config/configCmp/config.component.ts
```diff
@@ -185,4 +199,67 @@ export class ConfigComponent implements OnInit, OnDestroy {
...
+  public get showBackground() {
+    return this._showBackground
+  }
+
+  public set showBackground(value) {
+    this.nehubaViewer.config.layout.useNehubaPerspective.drawSubstrates.color[3] = value ? 0.2 : 0
+    // if (!value)
+    //   this.nehubaViewer.config.layout.useNehubaPerspective.drawSubstrates = false
+    // else
+    // this.nehubaViewer.config.layout.useNehubaPerspective.drawSubstrates.color[3] = 0.2
```
xgui3783, line 251: does this not work? If so, can you remove it?

src/ui/config/configCmp/config.component.ts
```diff
@@ -185,4 +199,67 @@ export class ConfigComponent implements OnInit, OnDestroy {
...
+
+    this.nehubaViewer.redraw()
+    this._showBackground = !this.sliceBackground || this.sliceBackground[3] > 0
+  }
+
+  public toggleShowBackground(value) {
+    this.showBackground = value
+  }
```
xgui3783, line 259: ditto previous comment. `toggleShowBackground` should not take any arguments. if it takes arguments like this, it should be called `setBackgroundVisibility`

Also `showBackground` is somewhat vague. I would add more context like `togglePerspectiveViewSubstrate` or s th

src/ui/config/configCmp/config.component.ts
```diff
@@ -185,4 +199,67 @@ export class ConfigComponent implements OnInit, OnDestroy {
...
+
+  private hexToRgb(hex) {
+    return hex.match(/\w\w/g).map(x => parseInt(x, 16))
+  }
```
xgui3783, line 263: can you add typing?

xgui3783 (changes requested):
Codewise, Just some minor changes. 

Thanks for the good work.

xgui3783:
also, it appears that the background colour changer is not working properly. 

at start, it sets to black

and on update, it will always reset to white.

Can you add some tests to ensure the methods works as intended?

I have a feeling that we want this in the gear icon mat menu, rather than in config. It's rather hard to find.

fsdavid:
> I have a feeling that we want this in the gear icon mat menu, rather than in config. It's rather hard to find.

To avoid overloading on the perspective views gear icon, which more feels a quick setting of the perspective view than all 4 views, I think using the `Settings` window is a good way. I also would like if users will get used to using the settings window, as it could hold more settings.

xgui3783:
> > I have a feeling that we want this in the gear icon mat menu, rather than in config. It's rather hard to find.
> 
> To avoid overloading of the perspective views gear icon, which more feels a quick setting of the perspective view than all 4 views, I think using the `Settings` window is a good way. I also would like if users will get used to using the settings window, as it could hold more settings.

I agree that perspective setting menu can become quite overloaded.

At the same time, I also think **at least** the substrate colour can be configured there, since it is a more natural place to try to find the setting. 

We can duplicate the UI in user setting menu (where it is currently), as well

PR #1157, opened 2022-02-07
Add hide/show parcellation button on the chip

fsdavid:


fsdavid:
I've commented on tests for hiding delineations. I will move them to the state if we agree about this approach.

xgui3783:
Hmm, I have a feeling that this is way more complicated than it needs to be. 

I don't think it needs to be on the global store. The more I think about this, the more I think using a directive to track, and act on the toggle might be a simpler approach.

fsdavid:
> Hmm, I have a feeling that this is way more complicated than it needs to be.
> 
> I don't think it needs to be on the global store. The more I think about this, the more I think using a directive to track, and act on the toggle might be a simpler approach.

What do you think about the service if you do not want to make the state heavier? if we use directive it should be adapted to the element (to change icon on toggle, also tooltip will not be dynamic). Current state parcellation is will not be available in the application.

fsdavid:
or maybe we can keep only `parcellationVisible` in the state?

fsdavid:
Feature is moved to the directive

xgui3783 (changes requested):
In general, good! This is what I had in mind when I asked you to change it to directive based.

src/viewerModule/nehuba/viewerCtrl/viewerCtrlCmp/viewerCtrlCmp.component.spec.ts
```diff
@@ -144,34 +144,34 @@ describe('> viewerCtrlCmp.component.ts', () => {
...
+        //   const slideToggle = await loader.getAllHarnesses(
+        //     MatSlideToggleHarness.with({
+        //       name: toggleName,
+        //     })
+        //   )
+        //   await slideToggle[0].toggle()
+        //   expect(
+        //     toggleDelination
+        //   ).toHaveBeenCalled()
+        // })
```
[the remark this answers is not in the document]
fsdavid: I will move it to the directive if you approve the current approach.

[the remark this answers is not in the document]
xgui3783: I very much approve this approach ^^

src/viewerModule/viewerStateBreadCrumb/toggle-parcellation.directive.ts
```diff
@@ -0,0 +1,86 @@
...
+    if (key === 'q' && target.classList.contains('neuroglancer-panel')) {
+      this.toggleParcellation()
+    }
+  }
+
+  @HostListener('click', ['$event'])
+  mouseclick(event) {
+    event.stopPropagation()
+    this.toggleParcellation()
+  }
```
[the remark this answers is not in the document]
fsdavid: Why don't we use a service then?

src/viewerModule/viewerStateBreadCrumb/toggle-parcellation.directive.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  toggleParcellation() {
+    this.renderer.removeClass(this.el.nativeElement, this.visible? 'fa-eye' : 'fa-eye-slash')
+    this.renderer.addClass(this.el.nativeElement, this.visible? 'fa-eye-slash' : 'fa-eye')
```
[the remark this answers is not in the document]
fsdavid: same as above

src/viewerModule/viewerStateBreadCrumb/toggle-parcellation.directive.ts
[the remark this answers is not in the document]
xgui3783: Hmm, my opinion is a service would have us need to inject the service in the component. This creates rather tight coupling between the host component and the service.

With a directive, the logic reside solely in the markup, and the component logic does not need to be complicated further.

xgui3783:
@fsdavid can you ping me when you have finished this PR and it is again ready for review?

Thanks for the work so far.

fsdavid:
> @fsdavid can you ping me when you have finished this PR and it is again ready for review?
> 
> Thanks for the work so far.

Sure :)

fsdavid:
@xgui3783 I have made some changes to the tests. 
I could not make a fix  for the `'> calls schedulRedraw'` and comment it

src/viewerModule/nehuba/viewerCtrl/viewerCtrlCmp/viewerCtrlCmp.component.spec.ts
```diff
@@ -144,34 +144,13 @@ describe('> viewerCtrlCmp.component.ts', () => {
       describe('> toggle delineation', () => {
         
         let toggleDelination: jasmine.Spy
-        const toggleName = 'toggle-delineation'
+        // const toggleName = 'toggle-delineation'
```
xgui3783, line 147: can you remove if not used, rather than comment?

src/viewerModule/viewerStateBreadCrumb/toggle-parcellation.directive.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  private subscriptions: Subscription[] = []
+
+  private nehubaInst: NehubaViewerUnit
+
+  get ngViewer() {
+    return this.nehubaInst?.nehubaViewer.ngviewer || (window as any).viewer
+  }
+  
+  constructor(private el: ElementRef,
```
xgui3783, line 31: I think `el` is not used anywhere?

src/viewerModule/viewerStateBreadCrumb/toggle-parcellation.directive.ts
```diff
@@ -0,0 +1,88 @@
...
+              private renderer: Renderer2,
```
xgui3783, line 32: neither is `renderer`

src/viewerModule/viewerStateBreadCrumb/toggle-parcellation.directive.ts
```diff
@@ -0,0 +1,88 @@
...
+              private pureService: PureContantService,
+              private store$: Store<any>,
+              @Optional() @Inject(NEHUBA_INSTANCE_INJTKN) private nehubaInst$: Observable<NehubaViewerUnit>,) {
+    this.subscriptions.push(
+      this.store$.select(viewerStateGetSelectedAtlas)
```
xgui3783, line 37: I did not know that store had a `.select` method natively. 

But still, let's keep the convention of using `.pipe(select(SELECTOR))`

src/viewerModule/viewerStateBreadCrumb/toggle-parcellation.directive.ts
```diff
@@ -0,0 +1,88 @@
+import {Directive, ElementRef, Inject, OnDestroy, Optional, Renderer2} from "@angular/core";
```
xgui3783, line 1: I would also rename the directive to `parcVisCtrl.directive.ts` in lieu with the change of the selector

src/viewerModule/viewerStateBreadCrumb/toggle-parcellation.directive.ts
```diff
@@ -0,0 +1,88 @@
...
+  viewerStateGetSelectedAtlas, viewerStateSelectedParcellationSelector,
+  viewerStateSelectedTemplatePureSelector
+} from "src/services/state/viewerState/selectors";
+import {select, Store} from "@ngrx/store";
+import {NehubaViewerUnit} from "src/viewerModule/nehuba";
+import {NEHUBA_INSTANCE_INJTKN} from "src/viewerModule/nehuba/util";
+
+@Directive({
+  selector: '[s-xplr-parc-vis-ctrl]',
+  exportAs: 'toggleParcellation'
```
xgui3783, line 14: I would export the directive as `sXplrParcVisCtrl`, in lieu with the updated selector

src/viewerModule/viewerStateBreadCrumb/toggle-parcellation.directive.ts
```diff
@@ -0,0 +1,88 @@
...
+  private selectedAtlasId: string
+  private selectedTemplateId: string
+
+  private subscriptions: Subscription[] = []
+
+  private nehubaInst: NehubaViewerUnit
+
+  get ngViewer() {
+    return this.nehubaInst?.nehubaViewer.ngviewer || (window as any).viewer
+  }
```
xgui3783, line 29: there should no longer a need for a getter like this. all reference to the getter can be pointed to `this.nehubaInst?.nehubaViewer.ngviewer`

xgui3783 (changes requested):
Can you also remove the commit you did for `package-lock.json`?

fsdavid:
@xgui3783 It should be fixed now

xgui3783:
Thanks. 

I will take a look maybe today, maybe tomorrow. 

I will also investigate why `scheduleRedraw` no longer gets called.

xgui3783:
superceded by #1169 , specifically 83b7609bfa772968e846c0c4377365dabd6e1ca0

PR #1189, opened 2022-05-04
Refactor migrate connectivity - change datasets with slider

fsdavid:


.storybook/preview.js
```diff
@@ -1,6 +1,7 @@
 import { setCompodocJson } from "@storybook/addon-docs/angular";
 import docJson from "../documentation.json";
 setCompodocJson(docJson);
+import * as ConnectivityComponent from 'hbp-connectivity-component/dist/loader'
```
xgui3783, line 4: Hmm, I wonder if we could use unpkg rather than include it as a dependency. 

see `src/index.html` and how `ng-layer-tune` is included

src/atlasComponents/sapi/core/sapiParcellation.ts
```diff
@@ -36,12 +36,13 @@ export class SAPIParcellation{
     )
   }
 
-  getFeatures(param?: PaginationQuery, queryParam?: SapiQueryParam): Observable<SapiParcellationFeatureModel[]> {
+  getFeatures(param?: PaginationQuery, type?: string, queryParam?: SapiQueryParam): Observable<SapiParcellationFeatureModel[]> {
```
xgui3783, line 39: I would create a new type, extending the PaginationQuery, called ParcellationPaginationQuery, which contains `type: string` as an optional param

src/atlasComponents/sapi/core/sapiParcellation.ts
```diff
@@ -36,12 +36,13 @@ export class SAPIParcellation{
...
     return this.sapi.httpGet<SapiParcellationFeatureModel[]>(
       `${this.sapi.bsEndpoint}/atlases/${encodeURIComponent(this.atlasId)}/parcellations/${encodeURIComponent(this.id)}/features`,
       {
-        per_page: '5',
-        page: '0',
+        type: type,
+        size: param && param.size? param.size.toString() : '5',
```
xgui3783, line 44: this can probably be:

`size: param?.size?.toString() || '5'`

src/atlasComponents/sapi/sapi.service.ts
```diff
@@ -81,6 +81,10 @@ export class SAPI{
     return reg.getFeatures(spaceId)
   }
 
+  getModalities() {
+    return this.http.get('https://siibra-api-dev.apps-dev.hbp.eu/v1_0/modalities')
+  }
+
```
xgui3783, line 87: do not hard code the URLs.

src/atlasComponents/sapiViews/core/region/module.ts
```diff
@@ -17,6 +17,7 @@ import { SapiViewsCoreRegionRegionRich } from "./region/rich/region.rich.compone
     SapiViewsUtilModule,
     SapiViewsFeaturesModule,
     SpinnerModule,
+    SapiViewsFeaturesModule,
```
xgui3783, line 20: already imported, this is not needed

src/atlasComponents/sapiViews/core/region/region/rich/region.rich.component.ts
```diff
@@ -36,12 +38,12 @@ export class SapiViewsCoreRegionRegionRich extends SapiViewsCoreRegionRegionBase
 
   // eslint-disable-next-line  @typescript-eslint/no-empty-function
   handleExpansionPanelClosedEv(title: string){
-    
+    this.expandedPanel = ''
```
xgui3783, line 41: makes more sense if `this.expandedPanel = null`

src/atlasComponents/sapiViews/core/region/region/rich/region.rich.template.html
```diff
@@ -99,8 +93,43 @@
...
+                                                   [accordionExpanded]="expandedPanel === CONST.CONNECTIVITY"
+    >
+    </sxplr-sapiviews-features-connectivity-browser>
+  </ng-template>
+
+  <ng-container *ngTemplateOutlet="ngMatAccordionTmpl; context: {
+    title: CONST.CONNECTIVITY,
+    iconClass: 'fab fa-connectdevelop',
+    content: sxplrSapiviewsFeaturesConnectivityBrowser,
+    contentAttribute: {types: hasConnectivityDirective.availableModalities, defaultProfile: hasConnectivityDirective.defaultProfile},
```
xgui3783, line 116: `contentAttribute` is not used anywhere? in which case remove it?

src/atlasComponents/sapiViews/core/region/region/rich/region.rich.template.html
```diff
@@ -148,3 +177,9 @@
     </ng-template>
   </mat-expansion-panel>
 </ng-template>
+
+<!-- fall back if region is not provided -->
+
+<ng-template [ngIf]="!region">
+  Region must be provided!
+</ng-template>
```
xgui3783, line 185: I think there is already a fallback near the header, and thus this can be removed

xgui3783 (changes requested):
Thanks for the PR. Mostly cosmetic changes in the code. 

Some additionall things:

- can you remove the place holder connectivity matrix component, as well as the stories associated with it
- connectivity component does not seem to play well in stories:
  - https://siibra-explorer.apps-dev.hbp.eu/refactormigrateconnectivity2/storybook-static/?path=/story/atlas-components-sapi-views-core-region-region-rich-region--injection-simple-region
  - https://siibra-explorer.apps-dev.hbp.eu/refactormigrateconnectivity2/storybook-static/?path=/story/atlas-components-sapi-views-features-connectivity-connectivity-browser-example-connectivity--default

can you take a look why?

.storybook/preview.js
fsdavid: I could not make it work with importing on `index.html` or `.storybook/preview-head.html`
Other issues are fixed

src/atlasComponents/sapi/core/sapiParcellation.ts
```diff
@@ -8,6 +8,10 @@ type PaginationQuery = {
   page: number
 }
 
+type ParcellationPaginationQuery = {
+  type?: string
+}
+
```
xgui3783, line 14: Hmm, I think what I meant originally is:

```ts
type ParcellationPaginationQuery = {
  type?: string
} & ParcellationPaginationQuery
```

so that in `getFeatures`, you will only need one additional argument.

src/atlasComponents/sapi/sapi.service.ts
```diff
@@ -21,9 +21,10 @@ type RegistryType = SAPIAtlas | SAPISpace | SAPIParcellation
 @Injectable()
 export class SAPI{
   static bsEndpoint = `https://siibra-api-dev.apps-dev.hbp.eu/v1_0`
+  static modalitiesEndpoint = `https://siibra-api-dev.apps-dev.hbp.eu/v1_0/modalities`
```
xgui3783, line 24: this is actually not what I meant. 

Imagine now that we change the URL, and it is no longer at `https://siibra-api-dev.apps-dev.hbp.eu`, but at `https://siibra-api.ebrains.eu`

The current implementation means that we will need to change the end point at two separate places, and prone to error. 

`SAPI.bsEndpoint` should be the only place where the hostname, version etc is referenced, and instances where the endpoint is used simply need to reference the `SAPI.bsEndpoint`

In our case, we can simply have:

```typescript
getModalities(){
  return this.http.get(`${SAPI.bsEndPoint}/modalities`)
}
```

src/atlasComponents/sapi/sapi.service.ts
```diff
@@ -82,7 +83,7 @@ export class SAPI{
   }
 
   getModalities() {
-    return this.http.get('https://siibra-api-dev.apps-dev.hbp.eu/v1_0/modalities')
+    return this.http.get(SAPI.modalitiesEndpoint)
```
xgui3783, line 86: ditto, please see above

src/atlasComponents/sapiViews/features/connectivity/connectivityBrowser/connectivityBrowser.stories.ts
```diff
@@ -57,11 +60,36 @@ class ExampleConnectivityBrowserWrapper {
...
   regionName: string = 'Area TE 3 (STG) right'
+  type: string = 'siibra/features/connectivity/streamlineCounts'
+  pageNumber = 1
+  numberOfDatasets = 1
   private regionIndexInMatrix = -1
   public connectionsString: string
 
 
-  constructor(private sapi: SAPI) {
+  constructor(private sapi: SAPI, private cdf: ChangeDetectorRef) {
```
xgui3783, line 70: cdf not used anywhere? 

Not overly concered, but good to know if it was injected via error

xgui3783 (changes requested):

xgui3783 (changes requested):
I __think__ you also need to import your component in `preview-head.js`

src/atlasComponents/sapi/core/sapiParcellation.ts
fsdavid: By adding the same type I'm getting an error `Type alias 'ParcellationPaginationQuery' circularly references itself.`

src/atlasComponents/sapiViews/features/connectivity/connectivityBrowser/connectivityBrowser.stories.ts
fsdavid: no, actually I missed adding manual detection of web component changes.

src/atlasComponents/sapi/sapi.service.ts
fsdavid: > preview-head.js

Sorry, you mean preview-head.html yes?

fsdavid:
> 

Sorry, you mean `preview-head.html` yes?

fsdavid:
Moved here: https://github.com/FZJ-INM1-BDA/siibra-explorer/pull/1192

Issue #1230, opened 2022-08-30
[UI] Atlas selection/ breadcrumb redesign


xgui3783:
Follow up from dev meeting on 2022 08 30, new breadcrumb/atlas selection UI will be exampled, mainly:

- use breadcrumb: <atlas> <parcellation> <space> <region>
- Remove atlas selector, atlas drop down

from https://github.com/FZJ-INM1-BDA/siibra-explorer/issues/1131

xgui3783:
fixed in 2.8.0

Issue #1190, opened 2022-05-11
Include Julich-Brain Atlas as parcellation map


SBludau:
In siibra python, Julich-Brain is a parcellation and MNI152 or colin27 is a space.

In siibra explorer inside the layer box the Julich-Brain atlas does not appear directly under parcellations but can only be selected as default parcellation via the spaces.

Shouldn't Julich-Brain also appear as a parcellation?

dickscheid:
This is a valid question. What I suggest is to add the cytoarchitectonic maps to the list of maps in the layer selector. Currently, they are not inlcueed there since they are considered the "base layer" which is always active. But it is true that this makes them less visible.

xgui3783:
related to #1230

xgui3783:
fixed in 2.8.0

PR #1238, opened 2022-09-12
New layer selector with chip style

fsdavid:


src/atlasComponents/sapiViews/core/atlas/smartChip/atlas.smartChip.components.ts
```diff
@@ -0,0 +1,36 @@
+import {Component, EventEmitter, Input, Output} from "@angular/core";
```
xgui3783, line 1: ditto see comment from previous PR

xgui3783 (commented):
You don't have to start a new branch/PR. 

if it is about the same feature, it is usually less confusing to commit on top of previous branch, so that previous comments are not lost.

src/atlasComponents/sapiViews/core/atlas/smartChip/atlas.smartChip.components.ts
fsdavid: Currently, all 3 components for atlas, space, and parcellation are different enough to keep them separate. especially selector for parcellation

fsdavid:
> You don't have to start a new branch/PR.
> 
> if it is about the same feature, it is usually less confusing to commit on top of previous branch, so that previous comments are not lost.

Sure, sorry for that. I do not know why, but the previous version was not building correctly. After rendering new commit, viewer was not updating, (cache clean was not working as well).

xgui3783:
@fsdavid can you take a look at the PR #1269 

specifically, a generic smart chip: https://siibra-explorer.apps-dev.hbp.eu/feataltatpselector/storybook-static/?path=/story/components-smartchip--default

how it looks as a component (color picker should work here): https://siibra-explorer.apps-dev.hbp.eu/feataltatpselector/storybook-static/?path=/story/atlascomponents-sapiviews-core-rich-atpselector-puredumb-pureatpselector--default

and how it interacts (e.g. check if template/parc is compatible etc) https://siibra-explorer.apps-dev.hbp.eu/feataltatpselector/storybook-static/?path=/story/atlascomponents-sapiviews-core-rich-atpselector-wrapper--default

I am not 100% happy with how the UI looks either, but it should be a good way to test out UIs before pushing them to dev/prod

Can you take a look at the PR, and ideally make your implementation as a storybook story, so that we can evaluate it in an upcoming dev  meeting?

xgui3783:
fixes #1230 #1190

PR #1242, opened 2022-09-20
Merge to dev

fsdavid:

PR #1257, opened 2022-11-08
Full view with perspective navigation

fsdavid:
Reviews are here: https://github.com/FZJ-INM1-BDA/siibra-explorer/pull/1242


xgui3783:
usability:

expected: maximisation of a panels hould not change the orthogonal view

actual: maximisation of a panel causes the perspective view to navigate to one of the orthogonal views

deploy/package-lock.json
```diff
@@ -1,8 +1,3736 @@
 {
   "name": "deploy",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
xgui3783, line 4: is there any good reason why we are updating the package-lock.json file in this commit?

src/layouts/layouts/single/single.template.html
```diff
@@ -1,3 +1,9 @@
-<div class="w-100 h-100">
+<div class="w-100 h-100 position-relative">
   <ng-content select="[cell-i]"></ng-content>
+
+  <div class="position-absolute"
+       style="bottom: 50px; right: 50px;">
```
xgui3783, line 5: use id + style, rather than inlining the style

I think there might even be a global 

Also, I wonder if using `fixed` rather than `absolute`, in which case, the containing block does not need to have `position: relative` attribute

src/services/maximiseView.service.ts
```diff
@@ -0,0 +1,301 @@
+import {select, Store} from "@ngrx/store";
```
xgui3783, line 1: it seems this service should be named as pictureInPicture.service.ts, no?

to me, it looked like this service negotiate between the pip navigation and single slice view in pip mode?

src/services/maximiseView.service.ts
```diff
@@ -0,0 +1,301 @@
...
+  private subscriptions: Subscription[] = []
+
+  public maximisedPanelIndex: number | null = null
+
+  public defaultZoomLevels
+
+  public navPosVoxel: any
+  public navPosReal: any
+  public defaultOrientation: any
+  public templateTransform = []
```
xgui3783, line 32: typing

src/services/maximiseView.service.ts
```diff
@@ -0,0 +1,301 @@
...
+
+  public isMaximised: boolean = false
+
+  private perspectivePanel: any
+
+  public max: any
+  public min: any
+  public height: any
+  public top: any
+  public topForSagittal: any
```
xgui3783, line 45: typing

src/services/maximiseView.service.ts
```diff
@@ -0,0 +1,301 @@
...
+        this.perspectivePanel.viewer.orthographicProjection.value = !clear
+
+        // Toggle scale bar
+        if ((this.perspectivePanel.viewer.showScaleBar.value && !clear)
+          || !this.perspectivePanel.viewer.showScaleBar.value && clear) {
+          this.perspectivePanel.viewer.showScaleBar.toggle()
+        }
+      }
+    }
+  }
```
xgui3783, line 235: why do we still need to update perspective view, now that we are using static images for pip?

src/services/maximiseView.service.ts
```diff
@@ -0,0 +1,301 @@
...
+const defaultZoom = {
+  "minds/core/referencespace/v1.0.0/dafcffc5-4826-4bf1-8ff6-46b8a31ff8e2": [3300000, 2350000, 1950000],
+  "minds/core/referencespace/v1.0.0/7f39f7be-445b-47c0-9791-e971c0b6d992": [2700000, 1850000, 1650000],
+  "minds/core/referencespace/v1.0.0/a1655b99-82f1-420f-a3c2-fe80fd4c8588": [1700000, 1750000, 1450000] ,
+  "minds/core/referencespace/v1.0.0/MEBRAINS": [3450000,2700000,1300000],
+  "minds/core/referencespace/v1.0.0/265d32a0-3d84-40a5-926f-bf89f68212b9": [165000,140000,100000],
+  "minds/core/referencespace/v1.0.0/d5717c4a-0fa1-46e6-918c-b8003069ade8": [600000,240000,240000],
+}
+
+export const perspectiveScreenshots = {
```
xgui3783, line 294: this does not need to be exported, since it's not used anywhere else?

src/services/maximiseView.service.ts
```diff
@@ -0,0 +1,301 @@
...
+
+              const topForSagittal = (viewSize-lineHeight)/2
+
+              this.max=max
+              this.min=min
+              this.height = lineHeight
+              this.top = top
+              this.topForSagittal = topForSagittal
+              
+              this.heightChanged.emit()
```
xgui3783, line 143: it seems this should be called navigation changed, not height changed

the name of the variable suggested to me that the viewport height had changed, but that is not the case.

src/viewerModule/nehuba/layoutOverlay/nehuba.layoutOverlay/nehuba.layoutOverlay.component.ts
```diff
@@ -111,22 +113,32 @@ export class NehubaLayoutOverlay implements OnDestroy{
 
   public volumeChunkLoading$: Subject<boolean> = new Subject()
 
+  public panelOrder: string
```
xgui3783, line 116: this variable is not used anywhere?

src/viewerModule/nehuba/layoutOverlay/nehuba.layoutOverlay/nehuba.layoutOverlay.component.ts
```diff
@@ -266,7 +278,7 @@ export class NehubaLayoutOverlay implements OnDestroy{
         }
         case "SINGLE_PANEL": {
           const element = removeExistingPanels()
-          const newEl = getSinglePanel(viewPanels)
+          const newEl = getSinglePanel(viewPanels, this.currentOrder)
```
xgui3783, line 281: ditto before. we should have a new layout called pip, rather than patching the current single layout

src/viewerModule/nehuba/layoutOverlay/nehuba.layoutOverlay/nehuba.layoutOverlay.style.css
```diff
@@ -39,3 +39,86 @@ current-layout
...
+  -webkit-appearance: none;
+  outline: none;
+  opacity: 1;
+  left: 0;
+  min-height: 5px;
+  background-color: rgba(0,0,0,0);
+  -webkit-transition: .2s;
+  transition: opacity .2s;
+  z-index: 200;
+  width:200px;
```
xgui3783, line 102: usually, I would say avoid using px as measuring units. 

use em or rem if you have to use absolute units.

src/viewerModule/nehuba/layoutOverlay/nehuba.layoutOverlay/nehuba.layoutOverlay.template.html
```diff
@@ -149,4 +152,45 @@
...
+</ng-template>
+
+<ng-template #miniPerspectiveTmp>
+  <div *ngIf="panelOrder && panelOrder[0] !== '3'" class="pe-all position-relative mini-perspective-viewer overflow-hidden">
+
+    <div *ngIf="maximiseService.maximisedPanelIndex !== null && maximiseService.templateTransform"
+         [style.background-image]="'url(' + maximiseService.previewImage + ')'"
+         style="background-size: cover;">
+
+      <input type="range"
```
xgui3783, line 164: why does there need to be two input components?

src/services/maximiseView.service.ts
fsdavid: I think it looks like `perspectiveViewMap` more, but anyways I will change the naming to pip.

xgui3783: but it is not active when we don't have a pip mode on?

perhaps perspectiveViewPip.service.ts?

fsdavid: I think it is not a pip as it is just a non-dynamic picture of perspective view. We could call it the `perspectiveViewSlider` (I've named separated component the samely).

xgui3783: pip does not necessarily mean it's dynamic, it is simply picture in picture. 

hmm, I am actually questioning the necessity of this service at all. 

couldn't this be done as a smart component?

fsdavid: Yes it could

src/viewerModule/nehuba/layoutOverlay/module.ts
```diff
@@ -10,6 +10,7 @@ import { WindowResizeModule } from "src/util/windowResize";
 import { LayoutModule } from "src/layouts/layout.module";
 import { MatButtonModule } from "@angular/material/button";
 import { MatTooltipModule } from "@angular/material/tooltip";
+import {AngularMaterialModule} from "src/sharedModules";
```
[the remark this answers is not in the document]
fsdavid: It uses material icon buttons.

src/viewerModule/nehuba/layoutOverlay/nehuba.layoutOverlay/nehuba.layoutOverlay.component.ts
fsdavid: It is used on template: https://github.com/FZJ-INM1-BDA/siibra-explorer/blob/9a8611b951da94acdc09aeed5142c4c63e77287b/src/viewerModule/nehuba/layoutOverlay/nehuba.layoutOverlay/nehuba.layoutOverlay.template.html#L158

src/viewerModule/nehuba/layoutOverlay/nehuba.layoutOverlay/nehuba.layoutOverlay.component.ts
fsdavid: Actually, getSinglePanel works for pip view as well. I will just copy paste the code in new `getPipPanel` method

src/viewerModule/nehuba/layoutOverlay/nehuba.layoutOverlay/nehuba.layoutOverlay.template.html
fsdavid: One of them represents zoom and position

src/viewerModule/nehuba/layoutOverlay/nehuba.layoutOverlay/nehuba.layoutOverlay.style.css
fsdavid: I may be wrong, but it is necessary to use pixels, as perspective view image is taken on specific position and size. If we make picture size responsive, it could display incorrect position. That is one of the reasons I've decided to use fixed units.

src/viewerModule/nehuba/layoutOverlay/module.ts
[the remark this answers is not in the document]
xgui3783: It should not?

src/viewerModule/nehuba/layoutOverlay/nehuba.layoutOverlay/nehuba.layoutOverlay.component.ts
xgui3783: If the two shares some logic, it makes sense to move the logic to a common method, rather than copy/paste code.

A good rule is, if a piece of code appear twice, there are generally something that can be done to improve it.

src/viewerModule/nehuba/layoutOverlay/nehuba.layoutOverlay/nehuba.layoutOverlay.style.css
xgui3783: IIRC, `rem` and `em` are simply `px`, but factors in high resolution displays (e.g. retina display in macos). 

I don't have a retina display, so I can't verify unfortunately. 

I guess we can wait until people complain before we fix it 😅

src/viewerModule/nehuba/layoutOverlay/module.ts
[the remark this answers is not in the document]
fsdavid: Why? :) little confused

src/viewerModule/nehuba/layoutOverlay/nehuba.layoutOverlay/nehuba.layoutOverlay.component.ts
fsdavid: Sure, I've changed it to simply use `getSinglePanel()` for both cases. we can rename it to get `fullViewPanel()`.

src/viewerModule/nehuba/layoutOverlay/nehuba.layoutOverlay/nehuba.layoutOverlay.template.html
xgui3783: hmm, I was sure that I also commented that the it maybe could be extracted into its own component. (something like a minimap component)

src/viewerModule/nehuba/layoutOverlay/nehuba.layoutOverlay/nehuba.layoutOverlay.style.css
fsdavid: I actually write it on retina display and it was not any problem :) . I just tried to make exact same size as a screenshot.

src/viewerModule/nehuba/layoutOverlay/module.ts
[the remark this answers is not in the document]
xgui3783: can you link where it uses mat-buttons? we can take a look?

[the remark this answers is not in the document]
fsdavid: Sure: https://github.com/FZJ-INM1-BDA/siibra-explorer/blob/9a8611b951da94acdc09aeed5142c4c63e77287b/src/viewerModule/nehuba/layoutOverlay/nehuba.layoutOverlay/nehuba.layoutOverlay.template.html#L84
These are viewer panel controllers.

src/viewerModule/nehuba/layoutOverlay/nehuba.layoutOverlay/nehuba.layoutOverlay.template.html
fsdavid: Yes, but I've separated the full <ng-template> into a different component named: `PerspectiveViewSlider`. (both inputs are together).

src/viewerModule/nehuba/layoutOverlay/nehuba.layoutOverlay/nehuba.layoutOverlay.component.ts
fsdavid: Ok, then I will create single method for `getSinglePanels` and `getPipPanels` and get unite logic there

src/viewerModule/nehuba/layoutOverlay/nehuba.layoutOverlay/nehuba.layoutOverlay.template.html
```diff
@@ -25,6 +25,9 @@
     [quick-tour-order]="quickTour3dViewSlide.order">
     <ng-content *ngTemplateOutlet="ngPanelOverlayTmpl; context: { panelIndex: currentOrder | getProperty : 3 | parseAsNumber }"></ng-content>
   </div>
+  <div *ngIf="panelOrder && panelOrder[0] !== '3'" class="w-100 h-100 position-relative" style="z-index: 100" picture-in-picture >
```
xgui3783, line 28: I know we talked about this briefly. I think it makes more sense to use `panelOrder` as an observable and use async pipe here. 

Or, in fact, more explicitly, in the component declaration, use `showPipPerspectiveView$`, rather than declaring `panelOrder` in component declaration, and hide the logic in template.

src/viewerModule/nehuba/viewerCtrl/change-perspective-orientation/changePerspectiveOrientation.component.html
```diff
@@ -1,40 +0,0 @@
-<button mat-menu-item [mat-menu-trigger-for]="perspectiveOrientationMenu">
```
xgui3783, line 1: hmm, I thought this PR does not affect this component?

You can checkout a specific version of the file with

```
git checkout -b <commitish> <path-to-file>
```

I think

fsdavid: As I remember I've added this component manually for this branch. Then I got the review that this is not related to the PR so I've removed it. I'm not sure it fill affect to original one, as the original should be created with another commit.

xgui3783: You can see in the diff that you are trying to remove the file that's already checked in.

src/viewerModule/nehuba/util.ts
```diff
@@ -27,7 +27,14 @@ const makeCol = (...els: HTMLElement[]) => {
...
 const washPanels = (panels: [HTMLElement, HTMLElement, HTMLElement, HTMLElement]) => {
   for (const panel of panels) {
-    if (panel) { panel.className = `position-relative` }
+    if (panel) {
+      panel.className = `position-relative`
+      panel.style.width = null
+      panel.style.height = null
+      panel.style.bottom = null
+      panel.style.right = null
+      panel.style.border=null
```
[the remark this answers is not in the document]
fsdavid: Yeah, as I see `washpanels` is not needed at all. I'm removing it/

src/viewerModule/nehuba/viewerCtrl/perspectiveViewSlider/perspectiveViewRangeValue.pipe.ts
```diff
@@ -0,0 +1,8 @@
+import { Pipe, PipeTransform } from "@angular/core";
+
+@Pipe({name: 'perspectiveViewRangeValue'})
```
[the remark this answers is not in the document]
fsdavid: What would you suggest?

xgui3783 (commented):
can you also fix the lint?

src/viewerModule/nehuba/viewerCtrl/perspectiveViewSlider/perspectiveViewRangeValue.pipe.ts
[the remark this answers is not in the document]
xgui3783: I think I can understand the code, but even then, I am not too sure what the pipe is trying to do.

e.g. I can understand panel probably should be `panelIdx`. What does value represent? It has any type, and it's not immediately clear what it is supposed to be.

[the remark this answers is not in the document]
fsdavid: It returns the position regarding the selected panel. 
image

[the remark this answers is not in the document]
xgui3783: that is the input, that is the `value` variable?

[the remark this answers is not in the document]
fsdavid: It is `navPosReal`, from where we need a single value, depending on the selected panel.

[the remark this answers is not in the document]
xgui3783: then name it as such. `value` is ambiguous, `navigationPositionReal: number[]` is easy to understand. 

the pipe, IMO, should be called `navigationPosToTextPipe`, since that's what the pipe is doing, taking navigation position, and convert it to text for output. 

I would also add `string` as the output of the pipe.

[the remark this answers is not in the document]
fsdavid: Done

src/viewerModule/nehuba/viewerCtrl/change-perspective-orientation/changePerspectiveOrientation.component.html
fsdavid: Actually, seems the coloring of vs code confused me, I've restored the files

src/layouts/layouts/pip/pip.style.css
```diff
@@ -0,0 +1,12 @@
+.major-column
+{
+  flex: 0 0 100%;
+}
+.minor-column
+{
+  flex: 0 0 0%;
+}
```
xgui3783, line 8: these are not needed?

src/viewerModule/nehuba/viewerCtrl/perspectiveViewSlider/navigationPosToTextPipe.pipe.ts
```diff
@@ -0,0 +1,8 @@
+import { Pipe, PipeTransform } from "@angular/core";
+
+@Pipe({name: 'perspectiveViewRangeValue'})
```
xgui3783, line 3: please also change pipe name

src/viewerModule/nehuba/viewerCtrl/perspectiveViewSlider/perspectiveViewSlider.component.ts
```diff
@@ -0,0 +1,227 @@
...
+@Component({
+  selector: 'nehuba-perspective-view-slider',
+  templateUrl: './perspectiveViewSlider.template.html',
+  styleUrls: ['./perspectiveViewSlider.style.css']
+})
+  
+export class PerspectiveViewSlider implements OnDestroy {
+
+    private nehubaViewer: any
+    private selectedTemplateId: string = ''
```
xgui3783, line 21: I think it makes much more sense to keep things as rxjs like as possible.

for example, I can imagine `selectedTemplateId` be replated by `selectedTemplate$` observable

src/viewerModule/nehuba/viewerCtrl/perspectiveViewSlider/perspectiveViewSlider.component.ts
```diff
@@ -0,0 +1,227 @@
...
+  
+    private subscriptions: Subscription[] = []
+  
+    public maximisedPanelIndex: number | null = null
+  
+    public navPosVoxel: number[]
+    public navPosReal: number[]
```
xgui3783, line 28: similarly, I think navPos can be an observable

xgui3783 (commented):
all in all really mature PR already.

Preferably, we can simplify perspectiveViewSlider component.

Lemme know if you have time so we can do some pair programming?

xgui3783:
@fsdavid , per our discussion, I have pushed a few changes to this PR. 

Can you take a look at them when you get a chance, and provide feedback?

xgui3783:
It appears to still have some buggy behaviours:

- tests are failing
- on init condition, it seems the minimap is disabled (errorenously)
- on template change, panel mode is not reset

fsdavid:
I think the zoom indicators for first and second slice views are calculated incorrectly, otherwise looks good and ready to merge.

fsdavid:
Sure

fsdavid:
I think current calculations do not work correctly, they depend on the screen size when the slice view does not. I could not understand the exact logic of how it calculates the size and margin of the position indicator now.

xgui3783:
Indicator should depend on viewport size IMO. 

From my experimentation, zoom seems to be defined as nm/px.

(e.g. does not matter which template we are in, if we call `viewer.navigationState.zoomFactor.restoreState(1)`, the scale bar will  show roughly show the same size. 

(e.g. for mni152 template, the finest resolution has resolution of [1e6, 1e6, 1e6] ` curl "https://neuroglancer.humanbrainproject.eu/precomputed/Fiber_Bundle/mni_icbm152_t1_tal_nlin_asym_09c_masked/info" | jq '.scales[0].resolution'` . When you load the template, then do `viewer.navigationState.zoomFactor.restoreState(1e4)`, you can see that the voxels are roughly 100 pixels across)

Knowing this, the viewport should influence the detail scrubber (red bar). 

Is there a particular position which are wildly wrong? Can you share a URL?

fsdavid:
It is not about the URL state, It is about screen size. For the first and second views, you can try different screen sizes and the indicator size will be different when it should not.

xgui3783:
> It is not about the URL state, It is about screen size. For the first and second views, you can try different screen sizes and the indicator size will be different when it should not.

Hmm, actually, the minor scrubber for slice view 1 and 2 are a bit wonky. I will investigate. It is likely that the same error contributes to both, since they both measure the superior-inferior zoom level.

In general though, should the size not be different when the viewport has different sizes? The visible area is different on resize. For example [this link](https://siibra-explorer.apps-dev.hbp.eu/devfullview/#/a:juelich:iav:atlas:v1.0.0:1/t:minds:core:referencespace:v1.0.0:dafcffc5-4826-4bf1-8ff6-46b8a31ff8e2/p:minds:core:parcellationatlas:v1.0.0:94c1125b-b87e-45e4-901c-00daee7f2579-290/@:0.0.0.-W000.._eCwg.2-FUe3._-s_W.2_evlu..7LIy..0.14gY0~.14gY0..rsg) when third panel is maximised, when viewport is 900px wide, it looks like:

image

But when the viewport is 400px wide, it looks like:

image

fsdavid:
For views 1 and 2, changing the window width should not affect the indicator, and it should be for view 3. The exact opposite for windows height - 1 and 2 should be affected and 3 not.

xgui3783:
@fsdavid you were right, panel 1 and 2 should depend on height, but was mistakenly made depend on width. 

It has been fixed in the latest commit.

xgui3783:
ping @fsdavid 

again, your intuition is correct. There is indeed a misalignment of the minimap scrubber, caused by center of the brain is not the same as the position 0, 0, 0

The latest commit should fix this issue. 

I also managed to fix the dimension issue for the 3rd slice view (previously, maximised 3rd slice view in big brain was rather wonky, but the latest commit fixes that too)

Once it builds, can you take a look and let me know if there

fsdavid:
I would say just 3 things about the zoom bar:
1 - It does not have a minimum width/height. When it is zoomed in, after some time it disapears
2 - For Allen mouse it is too small (All views)
3 - For the Waxholm it is also too small (Views 2 and 3)


Screenshot 2022-12-05 120620
Screenshot 2022-12-05 121200

xgui3783:
Good point. 

I will investigate why. 

I think they can be a 2.8.1 or 2.9 feature.

fsdavid:
> Good point.
> 
> I will investigate why.
> 
> I think they can be a 2.8.1 or 2.9 feature.

Ok, connectivity and multi-region selection fixes also will be added to  2.8.1

xgui3783:
@fsdavid I think I found the bug. Can you take a look at the PR again? (preview at https://siibra-explorer.apps-dev.hbp.eu/devfullview/#/ )

fsdavid:
@xgui3783 Sure, could you please explain what the bug is?

xgui3783:
> @xgui3783 Sure, could you please explain what the bug is?


@fsdavid 

The zoom mistakenly took width rather than height. Should have been fixed now.

fsdavid:
That is fixed. One issue is that if you maximize the view on the monkey brain and change the atlas to human, zoom and maximize icons are appearing incorrectly. Seems panel view sorting is not restarting.

xgui3783:
> One issue is that if you maximize the view on the monkey brain and change the atlas to human, zoom and maximize icons are appearing incorrectly. Seems panel view sorting is not restarting.

@fsdavid this seems to be a bug that exist in the current version of the viewer, and thus out of scope for this PR. 

We should raise an issue and deal with it in a separate PR.

fsdavid:
In this case, it works nicely

PR #1267, opened 2022-11-24
bugfix: setting custom layer overwrites exsting layers

xgui3783:
chore: added prettierrc


xgui3783:
@fsdavid I have made some changes to custom layers

instead of using queryparam, it now uses specific path (since it's more straight forward with the existing `customRoute` functionality.

now, by appending `/x-overlaylayer:${encodeUriComponent('nifti://http://localhost:3000/test.nii.gz')}`, one can overlay any volume on existing atlas

fsdavid:
Should tests be fixed before the merge?

xgui3783:
yes, I am looking into it

in the mean time, can you check the others?

fsdavid:
Sure

PR #1276, opened 2022-12-11
Add build artifacts

fsdavid:


xgui3783:
can you see if the build artefact can be done on release, and attach it as a release artefact?

fsdavid:
Actually, a build artifact cannot be attached as a release artifact. One solution could be to cache build and use it on the release workflow (But I'm not sure if it will work stable) or build it in the release workflow itself and upload it as a release asset. Second is that we can update the `readme` and include the URL of the artifact there. Unfortunately, I could not find a way to get the `latest` artifact. The third is to keep it as a built artifact, as it is now. I personally think the third solution is the best.

xgui3783:
superceded by #1279

PR #1278, opened 2022-12-19
Set default connectivity profile

fsdavid:


xgui3783:
looks good. 

can you add release notes? 

if you target dev, it would be 2.9.0, if you target staging, it would be 2.8.1

PR #1279, opened 2023-01-19
attach built artefacts as release asset

xgui3783:


fsdavid:
@xgui3783 is not `upload-release-asset` deprecated? I think GitHub actions API is a new way.
Also, from the 3 options in my comment if you decide to build an app in the release stage I think it is not a bad idea to build it in the testing stage as well.

fsdavid:
Oh, I see my comment came late.

xgui3783:
Not deprecated, just not actively maintained.

It is also GitHub action, and worst case scenario, we could always directly upload via curl

fsdavid:
Yeah, currently it works, I've also tested it, but what I found is that the repo is archived and does not maintain at all.
I just thought it would be the best scenario to use an actively maintainable approach (There are some forks that are actively maintained if there are any problems using curl), and also I thought it was not the best idea to build an app on the release workflow, that was the reason of my comment. But it's okay for me.

xgui3783:
> and also I thought it was not the best idea to build an app on the release workflow, that was the reason of my comment.

I actually have the exact opposite understanding. IMO, built artefacts are only useful at releases / tags. This is so that if the developers/users chooses to, they can easily roll back to an earlier version of the app. (This is also the reason why you can specify the exact version when one does `npm i` or `pip install` )

> ... I found is that the repo is archived and does not maintain at all. I just thought it would be the best scenario to use an actively maintainable approach ...

I see what you are saying. In fact the release action `actions/create-release@v1` is also archived and unmaintained. I am slightly wary of 3rd party actions, especially if first party actions can do the job, even if they are archived/unmaintained.

xgui3783:
> I see what you are saying. In fact the release action actions/create-release@v1 is also archived and unmaintained. I am slightly wary of 3rd party actions, especially if first party actions can do the job, even if they are archived/unmaintained.

This being said, I am immediately going to revert on what I said. softprops/action-gh-release@v1 looks mighty fine, with lots of bells and whistles. 

This might be a good opportunity to update to a more maintained action. 😅

fsdavid:
> I see what you are saying. In fact the release action `actions/create-release@v1` is also archived and unmaintained. I am slightly wary of 3rd party actions, especially if first party actions can do the job, even if they are archived/unmaintained.
I agree with that and found that the current official way to do it is the GitHub Actions API. I will look at the workflows and upgrade them if you are not against.

PR #1280, opened 2023-01-27
Add PWA (Make Siibra explorer installable)

fsdavid:


xgui3783:
Is this a requirement I missed?

I do not recall making siibra explorer pwa is a priority?

xgui3783:
is this regarding the mile stone `Allow local installation of the whole toolsuite` ?

This ain't' it, IMO. 

PWA only makes siibra explorer frontend installable, with no backend/data. The added baggage is that, now we needed to support an additional feature/point of failure. 

For the installation, since we need to ship siibra-api any way, siibra-explorer can be served by siibra-api.

fsdavid:
No, it is not about accessing templates locally. I've just enabled PWA to make the website installable for easy access, like google maps or youtube music. I believe `Allow local installation of the whole toolsuite` is a different feature, mostly connected to data. 

If `Allow local installation of the whole toolsuite` means creating Siibra explorer as an OS application I would happily implement it.

> The added baggage is that, now we needed to support an additional feature/point of failure.

Sorry, did not catch what could be the feature/point of failure. I belive PWA is more beneficial.

xgui3783:
> Sorry, did not catch what could be the feature/point of failure. I belive PWA is more beneficial.

Adding pwa will mean that there is one more feature we need to ensure that it doesn't break for each release. Especially for siibra explorer, this could effectively be achieved using a book mark.

xgui3783:
> With the feature, it is possible to install siibra explorer as a smartphone/tablet or PC app

But functionally, this is no different to `Add to homescreen` functionality on android (and probably something equivalent on safari on iphone. 

With the added downside of needing to maintain the additional feature.

fsdavid:
I may be wrong but `Add to homescreen` is just the shortcut that opens the browser with some specific URL, but with PWA application is a normal hybrid app, working on its window and Webview.

xgui3783:
> I may be wrong but Add to homescreen is just the shortcut that opens the browser with some specific URL, but with PWA application is a normal hybrid app, working on its window and Webview.

And what would be the benefit of using a webview over a regular browser window?

fsdavid:
I think there is a number of benefits to having a PWA application, mostly connected to comfortable use. Maybe, I'm wrong and it is comfortable only for me. I suggest showing it to the users.

xgui3783:
> I think there is a number of benefits to having a PWA application, mostly connected to comfortable use.

I failed to see that, especially at the cost of needing to maintain it.

It would prefer to not implement something unless there is a specific demand.

There's too much on our to-do list already.

xgui3783:
(I probably should be the last person to say stick to the plan..., But rewriting voblua, I have a reason, I swear!!!)

fsdavid:
Sure, I agree.

PR #1291, opened 2023-02-28
Fix connectivity issues

fsdavid:


xgui3783 (changes requested):
LGTM

just one thing, can you add the 2.9.1 entry in mkdocs.yml?

fsdavid:
Ah, I knew I was missing something. Thanks

PR #1294, opened 2023-03-27
v2.10.0

xgui3783:


xgui3783:
sane URL currently broken. 

need to relook into http priority interceptors

Issue #1336, opened 2023-06-01
allow custom neuroglancer shaders to be placed in a shader.json file


PolarBean:
Hi, 

apologies for all the requests, siibra is great and I am enjoying using it. Would it be possible in the future to allow the inclusion of a shader.json in the same manner as transform.json is included. This file would include the code you would normally find in the shader input box under the render tab in neuroglancer. 

Thanks in advance!

xgui3783:
This has been talked about with @pacher . 

I would argue that it should be called something more generic like `volume.info.json`.

It can contain the following fields:

- min
- max
- shader
- transform

for now. 

The transform field is duplicated from `transform.json`, but it would be nice to eventually have only one metadata file, rather than two

ping @pacher if you have comments/suggestions

xgui3783:
discussed with @pacher 

likely, we will use a filename called `meta` for metadata

we will produce a json schema and a few examples, then call for an RFC.

PolarBean:
This all seems very reasonable!

xgui3783:
closing this issue for now. 

any further discussion should be had in #1348

PR #1348, opened 2023-06-14
RFC meta spec

xgui3783:
see: #1336 

ping @pacher @polarbean

Here is the first draft. Please provide feedbacks if any. 

Once we are reasonably happy, I will convert the spec into json schema.


PolarBean:
This all looks good. I cant think of anything else to add. Let me know if you need testing

pacher:
1. I am not a fan of just top level `min` and `max`. It is too generic and not clear that it refers to intensity/pixel value. There could be other mins and maxs in the future. And the volume even does not have to be an image. And we might need those values per channel.
So I would recommend to add nested structure describing content/data/values (needs a good name). Something like
```json
{
  "version": 1,
  "data": {
    "type": "image", //optoinal, "image" by default and enough for now, could be more specific, like Microscopy/MRI or less specific
    "range": [{"min": 3,"max": 7}, {}, {"max": 25}] // only if "type" == "image", array for multi-channel volumes, object for single channel
  },
  "transform": ...
}
```
There should be some kind of semantic `type` of data, not the image/segmentation from precomputed format, not the dataType like int/float, but more like `intent` in nifti. The type will change which other properties this section has. For now it is enough to have one default like "image" and let it be optional.
There will be other useful properties of the image content/data to add. For example nehuba has `imageBackground?: vec3` and tiff has a tag `photometricInterpretation` with values like `BLACK_IS_ZERO` or `RGB`

2. About transform, just keep in mind that it is already in "legacy" format and make sure that it will not be painful to upgrade to the newer one with units. Also it should probably be described better (nanometer, points to the corner of the voxel..) 
3. Shader: too much prominent as it is. Right now it is pretty much in the center of attention like it wants to be used. With just a little remark. Not good. It is really a last resort hatch for hackers and should be buried under advanced/pro/force or whatever as deep as possible. There must be a way to request common (I mean everything we encounter) shaders by names and shortcuts.
Right now I need to display a volume which is RGB **AND** has a min/max range. I should not need to write a complex shader myself for that. And it should work with other potential future options or whatever else the viewer could provide like a brightness control. Shader is really "all bets are off, everything will be broken and that's on you" type of thing. It's a question if it should be included at all and we might try to remove it in future versions. So I would suggest something like
```json
{
  "override": {
    "customShader": ...
  }
}
```

pacher:
Not to mention that shader is code which could be broken in the future with no easy way to automatically update to a newer version. Shader does not really describe the data, so it has no place in metadata. We still want to have it for now to make it easier to move forward, but it should be clearly deprecated from the start.

xgui3783:
@pacher @PolarBean 

I have made some updates accodring to the feedbacks. Please note the following:

- I have created json schema file already, since it might actually be easier to generate documentation from the spec
- re: `range` property, I have currently opted to use an array, even for single channeled image, but using minItems maxItems to restrict number of items. IMO, this reduce the complexity of clients implementing parser

PolarBean:
I would like to note that the most valuable part of this update for me is the ability to add custom shaders to a volume. I understand that many users may not use this but I see no reason to hide the functionality behind some obscure layer. It also aligns nicely with the way neuroglancer functions :)

PolarBean:
Just following up to see if this is available for testing yet :)

xgui3783:
even though this PR is still outstanding, the feature is actually already rolled out on 2.12.3 (https://github.com/FZJ-INM1-BDA/siibra-explorer/pull/1382/files) when ng-layer-tune version got bumped to 0.0.18, al beit with a few bugs.

e.g.

```text
https://atlases.ebrains.eu/viewer/#/a:juelich:iav:atlas:v1.0.0:1/t:minds:core:referencespace:v1.0.0:a1655b99-82f1-420f-a3c2-fe80fd4c8588/p:minds:core:parcellationatlas:v1.0.0:94c1125b-b87e-45e4-901c-00daee7f2579-290/@:0.0.0.-W000.._eCwg.2-FUe3._-s_W.2_evlu..7eI0..1InsC.7aDN~.dAWb~..J4G/x-overlay-layer:precomputed:%2F%2Fhttps:%2F%2Fneuroglancer.humanbrainproject.eu%2Fprecomputed%2Fdata-repo%2FHSV-FOM
```

does not load the correct shader on initial load, but on refresh, it loads properly.

A fix (https://github.com/FZJ-INM1-BDA/siibra-explorer/pull/1389) has already been created, and will be deployed in the coming days. 

In the mean time, if you cannot wait, you can use https://atlases.ebrains.eu/viewer-staging/ to test your volumes.

i.e.

```
https://atlases.ebrains.eu/viewer-staging/#/a:juelich:iav:atlas:v1.0.0:1/t:minds:core:referencespace:v1.0.0:a1655b99-82f1-420f-a3c2-fe80fd4c8588/p:minds:core:parcellationatlas:v1.0.0:94c1125b-b87e-45e4-901c-00daee7f2579-290/@:0.0.0.-W000.._eCwg.2-FUe3._-s_W.2_evlu..7eI0..1InsC.7aDN~.dAWb~..J4G/x-overlay-layer:precomputed:%2F%2Fhttps:%2F%2Fneuroglancer.humanbrainproject.eu%2Fprecomputed%2Fdata-repo%2FHSV-FOM
```

(note the updated URL) should work properly (i.e. load the rgb color map on initial load)

incidentally

```bash
curl https://neuroglancer.humanbrainproject.eu/precomputed/data-repo/HSV-FOM/meta.json
```

is how the custom colormap (RGB) is fetched. 

I believe `override.customShader` would work similarly. 

but I must also stress Pavel's concern. Unless absolutely necessary/for debugging purposes,

PolarBean:
Ah I see, In the example posted it doesnt seem like you specify a colourmap in the meta.json, only that the volume is 3 channel. Is there a way to specify which colourmap you would like to use from the preset options (this is also a good way to avoid override.customShader)

PolarBean:
In the meta spec here: it says to use override.shader https://github.com/FZJ-INM1-BDA/siibra-explorer/blob/maint_shaderspec/.ngVolumeMetaSpec/meta.schema.v1.json 
However when i do it appears to be ignored. 

```python
{
    "$schema": "https://raw.githubusercontent.com/FZJ-INM1-BDA/siibra-explorer/maint_shaderspec/.ngVolumeMetaSpec/meta.schema.json",
    "version": 1,
    "data": {
        "type": "image/1d"
    },
    "override":{
        "shader": "void main() { emitGrayscale(toNormalized(getDataValue())); }"
    }
}
```

I will try changing this to override.customShader

PolarBean:
I think a nice feature to add to this spec would be a property for every option in  the volume tuning widget
image
allowing the volume creator to set default opacity, lower/higher threshold,  colormap, etc (perhaps this is already possible)

xgui3783:
>  lower/higher threshold,

We do intend to support this, via the `range` property

> hide clamped, colormap

We can certainly discuss in this PR how to best implement this

we are taking suggestions, if you have any, or I will add something to this PR, and we can discuss further.

PolarBean:
Reading through the spcec it seems specifying the meta version is required. I think its important to be explicit about where to fetch this information from and what effect it will have on how Siibra displays the volume. 
Also, I see that examples are included in the meta.schema.v1.json spec itself but this still requires a little bit of translation into what the actual meta file itself would look like. The inclusion of one or two functional example meta files would really improve the development experience. 

Other than that I fully support the changes proposed here and I look forward to testing them out once they are integrated into the live viewer.

xgui3783:
> Reading through the spcec it seems specifying the meta version is required

per https://github.com/FZJ-INM1-BDA/siibra-explorer/blob/maint_shaderspec/.ngVolumeMetaSpec/meta.schema.v1.json#L7 , this is expected to be a const for each version of meta. 

If we are lucky enough to see v2 of the meta schema, the version attribute will be a const with value `2`. (I would imagine, if there is a need to introduce any backwards incompatible changes, we will introduce v2 of meta. 

But indeed, I will include a desc on this attribute.

>  The inclusion of one or two functional example meta files would really improve the development experience.

There is currently one example meta file: https://github.com/FZJ-INM1-BDA/siibra-explorer/blob/maint_shaderspec/.ngVolumeMetaSpec/meta . 

the idea (originally, also currently) is to include a minimal example and one fully fleshed out one. 

I will get working to include more example meta file

PolarBean:
Great, a fully fleshed out example will be very useful. thank you.

xgui3783:
@PolarBean I have added a few more examples with the commit https://github.com/FZJ-INM1-BDA/siibra-explorer/commit/ea71b100ae9d027a931206f831f565577d3545b0

Are they what you expected?

PolarBean:
Very nice, perhaps clarify why multiple colormaps are specified?  
```json
   "preferredColormap": ["magma", "plasma", "greyscale"]
```
   Is this valid? or is it communicating to the user that they must select one of the following. 
     Another nice addition to the massive example would be the inclusion of override parameters (ie; how to include custom shaders).

Other than that this looks great!

PolarBean:
Hey :) has this been merged and available to use with siibra-explorer?

PolarBean:
Hey :) just wondering if this has been merged. 
I have the meta file at https://data-proxy.ebrains.eu/api/v1/buckets/d1-d2-expression/d1_ng/meta.json
However the viewer doesn't seem to like it nor request meta.json
[viewer link](https://atlases.ebrains.eu/viewer-staging/#/a:juelich:iav:atlas:v1.0.0:2/t:minds:core:referencespace:v1.0.0:265d32a0-3d84-40a5-926f-bf89f68212b9/p:minds:core:parcellationatlas:v1.0.0:05655b58-3b6f-49db-b285-64b5a0276f83/rn:f255401/@:0.0.0.-W000..2xhyO8.2-9sA4.-GQZj.2yO5eB..mJN..9CzQ~.3ksj.aF3~..5Bl/x-overlay-layer:precomputed:%2F%2Fhttps:%2F%2Fdata-proxy.ebrains.eu%2Fapi%2Fv1%2Fbuckets%2Fd1-d2-expression%2Fd1_ng)

xgui3783:
some update: 

It's quite silly on my part, but the code where meta is to be fetched is quite literally in, but just commented out:

https://github.com/FZJ-INM1-BDA/siibra-explorer/blob/master/src/atlasComponents/sapi/translateV3.ts#L571-L578

At the moment, we are quite swamped, however, and struggle to make a new release. I will make a commit that uncomment the code, but we can't promise when the new release of the viewer would be. 

One more thing, (mainly for my own record) I do not remember the result of the discussion whether we decided to use `/meta` or `/meta.json` (but not both), since I see examples of existence of both. 

since 

```
curl https://neuroglancer.humanbrainproject.eu/precomputed/data-repo/HSV-FOM/meta.json
```

exist, let's confirm once and for all that it is `/meta.json` that we wanted.

PolarBean:
/meta.json sounds good to me!

xgui3783:
@PolarBean 

Happy release day. Can you check [prod viewer link](https://atlases.ebrains.eu/viewer/#/a:juelich:iav:atlas:v1.0.0:2/t:minds:core:referencespace:v1.0.0:265d32a0-3d84-40a5-926f-bf89f68212b9/p:minds:core:parcellationatlas:v1.0.0:05655b58-3b6f-49db-b285-64b5a0276f83/rn:f255401/@:0.0.0.-W000..2xhyO8.2-9sA4.-GQZj.2yO5eB..mJN..9CzQ~.3ksj.aF3~..5Bl/x-overlay-layer:precomputed:%2F%2Fhttps:%2F%2Fdata-proxy.ebrains.eu%2Fapi%2Fv1%2Fbuckets%2Fd1-d2-expression%2Fd1_ng) displays the volume (including the metadata.json) as you expect?

.ngVolumeMetaSpec/meta.schema.v1.json
```diff
@@ -0,0 +1,279 @@
...
+                    "max": 0.8
+                },
+                {
+                    "min": 0.2
+                },
+                {
+                    "max": 0.8
+                }
+            ]
+        },
```
chourroutm, line 157: Of interest for https://github.com/FZJ-INM1-BDA/voluba/issues/198: does it mean display range or the boundaries of the histogram?

xgui3783: @chourroutm  at least when this is conceived, it signals to any viewer how it should define the boundary of histogram. In practice, we use this to define shaders, which would end up clipping values greater than max to max, and values lesser than min to min. 

I should double check the definitions, but I understand the value should respect the original data type (e.g. for uint16, 0 <= min <= max <= 65535 )

chourroutm: Great, thanks!

PR #1433, opened 2023-10-23
Add more meta for voi display

AhmetNSimsek:
- chenonceau
- wikibrainstem
- salditt
- broca/neun
- new pli mask


xgui3783 (approved):
Thanks

Issue #1458, opened 2024-02-26
Change "close" Option of Side Panel


KimLothmann:
When using the tool, Katrin always pressed the x to close the side panel (she wanted to collapse it). 

She thinks it would be better if the x was removed so that you don't accidentally close everything.

Suggestion: Add an "x close" at the bottom next to collapse?

![Collapse_Close_SidePanel](https://github.com/FZJ-INM1-BDA/siibra-explorer/assets/131377539/a3d2c922-0066-40d0-b25f-2b95daee4a81)

xgui3783:
an experimental fix has been implemented. we would appreciate any feedbacks.

To access the experimental fix, go to

https://siibra-explorer.apps.tc.humanbrainproject.eu/viewer-expmt/

and toggle the 📶 button to 🧪 at top right. 

You should see something similar to below:

![image](https://github.com/FZJ-INM1-BDA/siibra-explorer/assets/19381783/4dd1c1b3-8ffb-41b5-807e-adbf8651ab40)

PR #1460, opened 2024-02-28
experimental: region side panel

xgui3783:
feat: experimental update to selected region side panel
feat: allow experimental flag to be set at deploy time via env var
maint: update charts


xgui3783:
experimental fix to #1458

PR #1485, opened 2025-03-24
A few special treatment of configuration for Marmoset NM template

baishi:
added custom sacle for 3d model view of marmoset nm template
added special treatment of lightTheme for marmoset nm template
added special center translate for marmoset nm template
added custom substrate width and height for marmoset nm template


xgui3783:
Please note that it is on the roadmap that `meta.json` will eventually contain the necessary nehuba configuration (draw substrates, translate substrates, slice view zoom level), as well as theme (light & dark) in the near future.

